PR #53, opened 2016-01-12
Adding Ability to give custom URLs for plugins

tylerfontaine:
For custoemrs needing offline install of plugins, this adds the ability to use custom URLs.


gingerwizard:
@tylerfontaine will you be making another pull request here. Otherwise i'll address in the next release

PR #65, opened 2016-02-03
Fail if ERROR in stdout when installing plugin

robin13:
Otherwise many errors (e.g. just wrong plugin name) will be swallowed without plugin being installed successfully ...


gingerwizard:
@robin13 I believe the next pull request i'm adding will resolve this.  I'll move the discussion to the ticket as this failed tests.

PR #67, opened 2016-02-07
Added very basic support for marvel agent configuration

jakommo:
Added very basic support for marvel agent configuration. (Just exporters.type and exporters.host are currently supported)
@gingerwizard feel free to **not** merge this PR if you want to go with a more advanced solution. I just needed this for my home cluster anyways, so I thought I do a PR.


electrical:
I am against this since all config settings can and should be set via the config hash / map. Otherwise we have to maintain a large list of options which makes it impossible and ugly.

jakommo:
@electrical, good point, I agree. Closing this.

Issue #57, opened 2016-01-21
How do you add custom scripts/templates?


strootman:
The copy commands in the tasks/elasticsearch-scripts.yml and tasks/elasticsearch-templates.yml are pretty clear in their intent. 

So, is the expectation that I should clone this repo, and add any custom scripts/templates to the appropriate directory in `files/`?

Is there a way to configure a playbook to reference a different directory to load scripts/templates from?

gingerwizard:
Hi @strootman thanks for raising.  Currently a local template directory isn't configurable in the role so yes you would add to files/.

There will be a follow up release very shortly however and i agree this should be configurable - likewise for script.  Will add. 
Thanks

strootman:
:+1: Thank you

babadofar:
The same goes for other config files, dictionaries, synonyms, hunspell etc. Or should this be handled separately?

strootman:
I submitted a PR for this. Please take a look :)

gingerwizard:
@strootman apologies ive been out for a while. checking now. thanks.

gingerwizard:
@strootman merged. Thanks for the input! Happy to close?

strootman:
Yes, I think so.
Quick question though. This repo is the first I've seen that uses kitchen.ci. I was curious as to your thoughts about it.

I'm not sure if this is the best forum for having this discussion :)

gingerwizard:
Hi Strootman,
I'll close the ticket thanks.
Re kitchen.ci - it has soo far proved sufficient and has allowed us to ensure we have some repeatable tests across different platforms and versions.  The main issue we know face its limitation re being able to run a ansible role multiple times over a container/image  - thus preventing us from effectively testing that roles is idempotent as well as checking it upgrades plugins/versions appropriately.

I would love to hear peoples opinions on alternatives.  Please start a ticket to discuss possible testing strategies if your interested.

Thanks again for the pull request!

strootman:
Well, various `kitchen` commands take regex patterns. Maybe you could author a pair of _suites_ which are meant to be run together. Then you could execute with `kitchen test paired_test_suite*_[PLATFORM]`. The idea being that the second suite would re-run on the same host, cluster, node.

I haven't tried that out, but it seems possible.

gingerwizard:
[...]

PR #72, opened 2016-03-10
Custom scripts templates 57

strootman:
This PR addresses the enhancement described in Issue #57.

It uses the Ansible [with_fileglob](http://docs.ansible.com/ansible/playbooks_loops.html#id4) loop with a variable which can be set to a directory on the system which contains scripts or templates. If the variables are not set, then the behavior is unchanged.

I verified this by successfully running the `multi-2x-ubuntu-1404` test. I did not believe it necessary to add a test(s) for this since the relative globs succeeded in the test.


strootman:
I signed the CLA. Please let me know if I need to take further action.

gingerwizard:
LGTM. Thanks @strootman

PR #74, opened 2016-03-18
Fixes issue 70, Improves plugin management, Refactoring, Updates test…

gingerwizard:
This pull requests provides the following:

1. Improved plugin support. We now remove and re-install all plugins if the ES version changes (as required by Elasticsearch).  Otherwise we only re-install plugins if the user sets the appropriate flag.  Currently we don't dif the the installed vs the requested but this could be added. 
2. Fixes known issue with data dirs containing multiple paths and not all being modified.
3. Fixes known issues with init script and more than one node on a machine.
4. Refactors areas using with_items
5. Adds and fixes test issues.


gingerwizard:
jenkins, test it

gingerwizard:
jenkins, test it

gingerwizard:
jenkins, test it

Issue #36, opened 2015-12-15
Create user if it doesn't exist


gingerwizard:
Currently the default user "elasticsearch" is used - this is installed by the package manager.
If installed through a manual URL this might not be the case.  User may also wish to use another user.
Proposal is to create the user if it doesn't exist.

jerrac:
When I wrote my own ansible role to install elasticsearch, I eventually had to add configuration to set the elasticsearch users uid and gid. The reason being that mismatched uid/gid's on different nodes resulted in some nodes being unable to read/write snapshots. Since the repo directory was on an nfs mount with an owner/group of the elasticsearch user on the node I initiated the repo from. At I think that's the initial problem I ran into, it's been a while. 

I do know forcing all nodes to have the same username and uid/gid's fixed that issue.

So I'd love to be able to force the uid/gid's to be the same.

Of course, I haven't checked to see if the owner/group issue still occurs with elasticsearch 2.x. I suppose I'm posting this a bit early... :\

gingerwizard:
I'd like to get @electrical view on this. Easily added but its whether this is the appropriate place or whether it should be supported elsewhere.

jakommo:
I think that's only a problem when using `NFS`. S3 shouldn't have a problem with different ID's.

jerrac:
Yes, S3 would not have issues with id's. But we're not using it for this.

gingerwizard:
I can manually reproduce @jerrac issue.  Its hard for us to automate and test for this in the kitchen tests as we currently don't form clusters across docker images.  

The alternative is to ask users to create users using a separate ansible role and simply specify the es_user and es_group.  These will be used for the service.

jerrac:
If I remember correctly, the package manager should use the existing user if one exists, right? So if I create a "elasticsearch" user before installing elasticsearch, the package manager would use the user I created.

So, could we add a couple optional tasks to the role? If ```es_custom_user_id``` is ```True```, then create a user and a group named "elasticsearch" with the user and group ids in the ```es_custom_user_uid``` and ```es_custom_user_gid``` variables. After that, run the installation tasks.

Does that make sense?

jerrac:
So, what I described does seem to work just fine. See https://github.com/LaneCommunityCollege/ansible-elasticsearch/commit/de82918584edd5159643b43c1dff07cc4e5d63ca

Is that enough for a pull request? Or is there something else I need to do?

gingerwizard:
Hi
[...]

PR #75, opened 2016-03-18
For  #36 - Add support for creating optional elasticsearch user and group

jerrac:
Here's the pull request. :) Per https://github.com/elastic/ansible-elasticsearch/issues/36#issuecomment-198418652

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,10 @@ If installing Marvel or Watcher, ensure the license plugin is also specified.  S
 
 * ```es_user``` - defaults to elasticsearch.
 * ```es_group``` - defaults to elasticsearch.
+* ```es_user_id``` - default is undefined.
+* ```es_group_id``` - default is undefined.
+
+Both ```es_user_id``` and ```es_group_id``` must be set for the user and group ids to be set. 
 
 By default, each node on a host will be installed to use unique pid, plugin, work, data and log directories.  These directories are created, using the instance and host name, beneath default locations ]
 controlled by the following parameters:
--- a/tasks/elasticsearch-optional-user.yml
+++ b/tasks/elasticsearch-optional-user.yml
@@ -1,0 +1,18 @@
+---
+#Add the elasticsearch user before installing from packages.
+- name: Ensure optional elasticsearch group is created with the correct id.
+  group:
+    state: present
+    name: "{{ es_group }}"
+    system: yes
+    gid: "{{ es_group_id }}"
+
+- name: Ensure optional elasticsearch user is created with the correct id.
+  user:
+    state: present
+    name: "{{ es_user }}"
+    comment: elasticsearch system user
+    system: yes
+    createhome: no
+    uid: "{{ es_user_id }}"
+    group: "{{ es_group }}"
--- a/tasks/elasticsearch.yml
+++ b/tasks/elasticsearch.yml
@@ -8,6 +8,10 @@
 
 - debug: msg="Node configuration {{ es_config }} "
 
+- name: Include optional user and group creation. 
+  when: (es_user_id is defined) and (es_group_id is defined)
+  include: elasticsearch-optional-user.yml
+
 #- name: Include specific Elasticsearch
 #  include: "elasticsearch-{{ansible_os_family}}.yml"
 
--- a/test/integration/config.yml
+++ b/test/integration/config.yml
@@ -4,7 +4,7 @@
   hosts: localhost
   roles:
     #expand to all available parameters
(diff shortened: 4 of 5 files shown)


gingerwizard:
@jerrac apologies some other build issues associated with https://github.com/neillturner/kitchen-ansible/issues/7 resolving. Will test this once we resolve.

gingerwizard:
@jerrac Can we simplify this to just be es_user_uid & es_group_id? If either are specified we apply them to the elasticsearch user?

Also i think its worth raising an issue to be able to specify the user name i.e. change it from elasticsearch.  I suggest we assume this user exists - relying on other roles/playbook to create it.

jerrac:
@gingerwizard I'll switch to es_user_id and es_group_id.

> If either are specified we apply them to the elasticsearch user?

So, get ride of ```es_optional_user: True|False```, and just use ```es_user_id is defined or es_group_id is defined``` on the appropriate tasks?

> Also i think its worth raising an issue to be able to specify the user name i.e. change it from elasticsearch. I suggest we assume this user exists - relying on other roles/playbook to create it.

Don't we already have that? ```es_user```, and ```es_group``` in defaults/main.yml.

gingerwizard:
@jerrac yes we do! thanks :) I was confused because your pull uses literal "elasticsearch" rather than es_user and es_group.

Agreed re the other changes

gingerwizard:
@jerrac LGTM.  I'll test it shortly.

I think its worth adding a quick server spec test. Maybe to the config-2x and config-1x suites.  Are you happy to do this or shall i add?

gingerwizard:
jenkins, test it

jerrac:
I haven't used kitchen yet. I remember hearing about it once and thinking I should learn it. Heh. 

So, would adding tests for this just involve adding the vars to one of the playbooks in the test/integration directory?

gingerwizard:
@jerrac The tests seemed to have failed. It appears you need explicit is defined checks. It also worth checking the parameters are defined and not empty.

Yes ideally a test would use the parameters in one of the test playbooks - probably [config](https://github.com/elastic/ansible-elasticsearch/blob/master/test/integration/config.yml) is best.  Server spec can then be used confirm the user id and group id were set correctly after the role had been applied [here](https://github.com/elastic/ansible-elasticsearch/blob/master/test/integration/helpers/serverspec/config_spec.rb).

jerrac:
I don't need to edit the config_spec.rb files, right?

gingerwizard:
Yes you can add it there. It will then run for be tested for 1x and 2x. 
http://serverspec.org/resource_types.html#user
Is probably what you want.

jerrac:
Would this work for config-2x/serverspec/default_spec.rb? (Since those are the values I set in the config.yml file.)

```ruby
require 'config_spec'
require 'user_spec'
require 'group_spec'

describe 'Config Tests v 2.x' do
  include_examples 'config::init', "2.1.0"
end

describe group('elasticuser') do
  it { should exist }
end

describe group('elasticuser') do
  it { should have_gid 333 }
end

describe user('elasticuser') do
  it { should exist }
end

describe user('elasticuser') do
  it { should belong_to_group 'elasticuser' }
end

describe user('elasticuser') do
  it { should have_uid 333 }
end
```

btw, thanks for coaching me through this stuff. :smile:

jerrac:
Or maybe I should stick those in [config_spec.rb](https://github.com/elastic/ansible-elasticsearch/blob/master/test/integration/helpers/serverspec/config_spec.rb).

Speaking of that file, would this [section](https://github.com/elastic/ansible-elasticsearch/blob/master/test/integration/helpers/serverspec/config_spec.rb#L38) cause the tests to fail if the user is not _elasticsearch_? 

If so, should I keep the ```es_user```, and ```es_group``` values as "elasticsearch" in the test?

gingerwizard:
Hi
Put it in config_sepc and then it will run for 1.x and 2.x.  You can drop the 2.x test then.  I would keep the username as elasticsearch - we currently dont have code to create an alternative user just set the uid and guid.

gingerwizard:
Hey @jerrac thanks for this. We will test this, we're just experiencing some issues with the build env not clearing up docker images properly.  Once we've resolved this ill run it and hopefully merged. Please feel free to contribute elsewhere, its really appreciated.

gingerwizard:
infra problems hopefully solved so without further ado... 

jenkins, test it

gingerwizard:
@jerrac in an effort to improve stability the role has changed which is now causing conflicts. Can you please resolve and repush?

jerrac:
@gingerwizard Erm, did those changes help? Or am I missing something?

dliappis:
jenkins, test it

dliappis:
jenkins, test it

jerrac:
I rebased to master. Did that help? 

And after poking around Jenkins some more, I found where to view the errors that caused the failure. Which is what I was missing when I was checking things earlier. :\

gingerwizard:
jenkins, test it

gingerwizard:
@jerrac all tests pass! merging. Thanks for the contribution!

PR #76, opened 2016-03-18
Plugin Management and fixes

gingerwizard:
This pull requests provides the following:
1. Improved plugin support. We now remove and re-install all plugins if the ES version changes (as required by Elasticsearch). Otherwise we only re-install plugins if the user sets the appropriate flag. 
2. Currently we don't dif the the installed vs the requested but this could be added.
3. Fixes known issue with data dirs containing multiple paths and not all being modified.
4. Fixes known issues with init script and more than one node on a machine.
5. Refactors areas using with_items
6. Adds and fixes test issues.
7. Fix for Kitchen CLI issue

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -8,9 +8,9 @@ provisioner:
   roles_path: ../
   require_ansible_repo: true
   ansible_verbose: true
-  http_proxy: <%= ENV['HTTP_PROXY'] %>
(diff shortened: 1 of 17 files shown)


gingerwizard:
jenkins, test it

gingerwizard:
jenkins, test it

PR #95, opened 2016-04-15
ELS < 1.0 should start systemd instance in foreground

nerzhul:
Old instances have a -f flag to run in foreground mode. 

Without this flag, the instance is launched with shell & operator then systemd thinks instance doesn't work and kill the process.


gingerwizard:
@nerzhul thanks for the pull request. Unfortunately we don't support ES versions < 1.0.  The test matrix doesn't include tests and this would be prohibitively costly.

https://www.elastic.co/support/matrix

PR #98, opened 2016-04-18
Make an example in the README more readable.

winder:
One of the examples didn't have any formatting and was hard to read.


karmi:
Hi @winder, the CLA check is failing, maybe because you have committed with yout company mail, but signed with the Gmail e-mail. Can you please add both to your Github profile, and comment here, that should trigger a re-check?

winder:
There we go, you're right it was the email.

PR #100, opened 2016-04-20
added force option for package install, enabling users to downgrade

internatt:
use of the force=yes option here could be undesirable, but it does enable users to freely switch versions of elasticsearch on playbook runs.

### DIFF
--- a/tasks/elasticsearch-Debian.yml
+++ b/tasks/elasticsearch-Debian.yml
@@ -8,7 +8,7 @@
   when: es_use_repository
 
 - name: Debian - Ensure elasticsearch is installed
-  apt: name=elasticsearch{% if es_version is defined and es_version != "" %}={{ es_version }}{% endif %} state=present cache_valid_time=86400
+  apt: name=elasticsearch{% if es_version is defined and es_version != "" %}={{ es_version }}{% endif %} state=present force=yes cache_valid_time=86400
   when: es_use_repository
   register: elasticsearch_install
 


internatt:
Just signed the CLA. Jenkins is failing on the Centos7 build, appears to be a script issue with kitchen cleanup.

gingerwizard:
@dliappis we discussed this not failing the build.  Seems the change made hasn't had the desirable effect.

gingerwizard:
Also thoughts on force option @dliappis

dliappis:
@gingerwizard I added the changes to the post build section only for the Ansible Commit project to test its effect. I added it now to Ansible PR

dliappis:
jenkins, test it

dliappis:
@gingerwizard re: `force`: it seems that this is something required by the ansible apt module to allow downgrade (see: https://github.com/ansible/ansible-modules-core/pull/53). Under normal conditions `apt-get install packagename=version` does the trick.
 
My more general reservation is that offering downgrades is a tricky path to follow and actively support. What about plugins for example?
Additionally, if downgrade stays as a feature, equivalent support for yum (`yum downgrade`) needs to be added + some tests would need to be added to ensure it works properly too.

karmi:
Hi @internatt, the CLA check is still failing though we have your signature in our records. Maybe that's because you have used a different e-mail when signing the CLA and a when committing -- can you add both of those into your Github account and comment here? That should re-trigger the check.

internatt:
@karmi yes, the email in my .gitconfig was a work address for this commit. That's now been added to my github account.

@dliappis Agreed RE: supporting it as an option. As my commit description mentioned it may be undesirable, however I did find myself requiring this while prototyping and regression testing a cluster on AWS quite frequently.

jakommo:
I'm -1 on enabling `force` by default. It might set the wrong expectations regarding support for downgrades, which can lead to a lot of troubles.
I'd be ok with having this optional as a configuration option for testing though, if it is clear that this can break  stuff and should only by used by someone who knows what he is doing.

gingerwizard:
@internatt thanks for this and apologies for the late reply. Could you make the force option to default to false.  If the user sets allow_downgrades may we set? @jakommo

gingerwizard:
@internatt i'm going to add a force option now that defaults to no by adding a allow_downgrades flag.  I'll also update the documentation to reflect it can be configured but its for expert use only.  I'll reference the pull request here and close. Thanks again for the suggestion and pull request.

gingerwizard:
Note this will be supported on apt only. Ansible currently does not support 'yum downgrade' - see https://github.com/ansible/ansible-modules-core/issues/1419

gingerwizard:
Resolved through https://github.com/elastic/ansible-elasticsearch/pull/123

PR #102, opened 2016-04-26
Allow for selective command line task execution

strootman:
Adding tags to the `tasks/main.yml` allows the user to skip whole sets of tasks from the command line.

```
% ansible-playbook ... ansible-elasticsearch-playbook.yaml --list-tasks
...
    tasks:
      cyverse.elasticsearch : check-parameters	TAGS: [check]
      cyverse.elasticsearch : os-specific vars	TAGS: []
      include	TAGS: [java]
      include	TAGS: [install]
      include	TAGS: [config]
      include	TAGS: [scripts]
      include	TAGS: [plugins]
      include	TAGS: [service]
      include	TAGS: [templates]
```

```
% ansible-playbook ... ansible-elasticsearch-playbook.yaml --skip-tags "check,java,install" --list-tasks
...
    tasks:
      cyverse.elasticsearch : os-specific vars	TAGS: []
      include	TAGS: [config]
      include	TAGS: [scripts]
      include	TAGS: [plugins]
      include	TAGS: [service]
      include	TAGS: [templates]
```

Note: When you use `--list-tasks` with `--tags ...`, `include:` directives will still be listed. However, when you execute the playbook, only tasks which are marked with the selected tags will be executed.

# Example
So, if you only wanted to update some plugins, you would execute the following command;
```
% ansible-playbook ... ansible-elasticsearch-playbook.yaml --tags "plugins" 
```

# Other Details
Due to this role's use of the `set_fact` module, many of these had to have the `- always` tag applied to ensure that these tasks are always executed, thus preventing errors with conditional task execution (i.e. `when` clauses can/will break).

# Testing
No tests were added. I've tested this with the example listed above (`--tags "plugins"`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ The simplest configuration therefore consists of:
 
 The above installs a single node 'node1' on the hosts 'localhost'.
 
+This role also uses [Ansible tags](http://docs.ansible.com/ansible/playbooks_tags.html). Run your playbook with the `--list-tasks` flag for more information.
+
 ### Basic Elasticsearch Configuration
 
 All Elasticsearch configuration parameters are supported.  This is achieved using a configuration map parameter 'es_config' which is serialized into the elasticsearch.yml file.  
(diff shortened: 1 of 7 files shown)


strootman:
I checked the errors, and they are docker-related.

`bind: address already in use`

strootman:
@geoffroya, you may be interested in this.

gingerwizard:
@strootman apologies for the late reply. LGTM. Merging. Thanks for the commit.

Issue #104, opened 2016-05-10
data_dirs environment wrongly expanded


smalenfant:
Somehow this showed up after running the playbook in the `/etc/system/multi-user.target.wants/node1_elasticsearch.service` file 

`Environment=DATA_DIR=[,',/,v,a,r,/,l,i,b,/,e,l,a,s,t,i,c,s,e,a,r,c,h,/,e,s,-,0,1,-,n,o,d,e,1,',]`

Also shows in `/etc/elasticsearch/node1/elasticsearch.yml `
`path.data: [,',/,v,a,r,/,l,i,b,/,e,l,a,s,t,i,c,s,e,a,r,c,h,/,e,s,-,0,1,-,n,o,d,e,1,',]`

Expect : 

Template

```
[Service]
Environment=ES_HOME={{es_home}}
Environment=CONF_DIR={{conf_dir}}
Environment=DATA_DIR={{ data_dirs | array_to_str }}
Environment=LOG_DIR={{log_dir}}
Environment=PID_DIR={{pid_dir}}
EnvironmentFile=-{{instance_default_file}}
```

gingerwizard:
What version of Ansible are you using? This was a known issue with Ansible < 2.0 where variables would not retain their type when passed to Jinja

jakommo:
No feedback on the ansible version used. 
Closing for now.

PR #105, opened 2016-05-10
Removed array_to_string in 2 places

smalenfant:
Configuration files were not created properly. Not sure how that could have worked in the past, interested to know what I could be doing wrong.

### DIFF
--- a/templates/elasticsearch.yml.j2
+++ b/templates/elasticsearch.yml.j2
@@ -16,7 +16,7 @@ node.name: {{inventory_hostname}}-{{es_instance_name}}
 # Path to directory containing configuration (this file and logging.yml):
 path.conf: {{ conf_dir }}
 
-path.data: {{ data_dirs | array_to_str }}
+path.data: {{ data_dirs }}
 
 path.work: {{ work_dir }}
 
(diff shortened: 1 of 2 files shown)


smalenfant:
Fixes #104

gingerwizard:
As expected this has failed tests.  Are you using Ansible 2.0 as required?
This assumes the array is passed and not a string.  Prior to 2.0 it would be passed as a string.

smalenfant:
I believe this works with both now. But let me double check as I am using
the vagrant ansible provisioner. I'll also check the failed tests.

gingerwizard:
@smalenfant I satisfied this works with 2.0 or do you want to modify the pull request?

smalenfant:
Maybe the problem is related on how the hosts are specified. Here's how I used it..

`discovery.zen.ping.unicast.hosts: "host1:9300, host2:9300, host3:9300"`

Normally I think this should be a list.

gingerwizard:
@smalenfant path.data is handled differently from discovery.zen.ping.unicast.hosts.  The former you can declare as a list or string.  The latter is passed within the config map so will just be serialised as either a string or array - depending on how you pass it.  Either should work.

The array_to_str only works on 2.0 ansible.  Removing causes all tests to fail so i can't accept this pull request. Please open another issue if you are able to reproduce.

PR #109, opened 2016-05-25
fix for ES plugins install

dekhtyarev:
When using the ES plugins and restart(retry) the deployment, provided: **es_plugins_reinstall: false** we received an error when installing plugins, as their installation is not restricted (if defined).

I corrected this behavior, adding to check for **es_plugins_reinstall**.
But in this case we did not work the installation of plugins due to a bug in the variable **elasticsearch_install**.

Variable **elasticsearch_install** installation was recorded twice (feature ansible).
When you install from the repositories - we had the wrong variable **elasticsearch_install.changed**.

**elasticsearch_install** was divided into two different variables to correct this error.


dekhtyarev:
Build failed on:
> TASK [elasticsearch : Copy templates to elasticsearch]

I did not change variables for this task.
Problems with ansible version on jenkins or am I wrong? 

@dliappis can you help me in this matter?

tasks/elasticsearch-Debian.yml
```diff
@@ -19,4 +19,4 @@
 - name: Debian - Ensure elasticsearch is installed from downloaded package
   apt: deb=/tmp/elasticsearch-{{ es_version }}.deb
   when: not es_use_repository
-  register: elasticsearch_install
\ No newline at end of file
+  register: elasticsearch_install_from_package
```
antonbormotov: Please, add new line at the end

tasks/elasticsearch-RedHat.yml
```diff
@@ -14,9 +14,9 @@
...
   when: es_use_repository
-  register: elasticsearch_install
+  register: elasticsearch_install_from_repo
 
 - name: RedHat - Install Elasticsearch from url
   yum: name={% if es_custom_package_url is defined %}{{ es_custom_package_url }}{% else %}{{ es_package_url }}-{{ es_version }}.noarch.rpm{% endif %} state=present
   when: not es_use_repository
-  register: elasticsearch_install
\ No newline at end of file
+  register: elasticsearch_install_from_package
```
antonbormotov: +same

tasks/elasticsearch-Debian.yml
dekhtyarev: added.

tasks/elasticsearch-RedHat.yml
dekhtyarev: added.

bbaassssiiee:
This makes the role idempotent, great!

gingerwizard:
@dekhtyarev LGTM.
I will fix the test fail issue now - i'm able to reproduce locally.  I will then merge. Apologies for the delayed response.

gingerwizard:
@dekhtyarev merged in the changes through https://github.com/elastic/ansible-elasticsearch/pull/121 thanks for contribution!

Issue #99, opened 2016-04-20
Unable to set java version


internatt:
I've been testing the role with a few plugins that require java 8 (openjdk-8-jre-headless). Attempting to set the `java:` variable directly in the playbook, or in a separate vars file does not override the default. As a workaround, I can set it using the --extra-vars flag on runtime.

The use of `include_vars:` in `tasks/main.yml` is the culprit. According to [https://github.com/ansible/ansible/issues/11807](url) , this is 'expected behaviour'.

marji:
I have created [PR112](https://github.com/elastic/ansible-elasticsearch/pull/112) which enables to specify which java to install (by passing the java package name to the role).

Example: 
```
roles:
  - { role: elasticsearch, es_java: "openjdk-8-jre-headless", es_instance_name: "node1" }    
```
The role (current master as of June 5, 2016) fails on ubuntu 16.04 as it has hardcoded java package "openjdk-7-jre-headless" which is not present in ubuntu 16.04 repos.

gingerwizard:
@internatt  i believe we merged in a fix for this. Thanks for the resolution @marji

PR #112, opened 2016-06-05
enable specify java version (package name)

marji:
If **es_java** variable is set, it will use that as the java package name
instead of the default package.
This can be handy for example when running on Ubuntu 16.04 (which is not yet officially supported).

Example:
```
roles:
  - { role: elasticsearch, es_java: "openjdk-8-jre-headless", es_instance_name: "node1" }
```
The specified package must be available in the currently enabled repositories.

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -1,4 +1,4 @@
 ---
-java: "openjdk-7-jre-headless"
+java: "{{ es_java | default('openjdk-7-jre-headless') }}"
 default_file: "/etc/default/elasticsearch"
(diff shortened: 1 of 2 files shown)


marji:
This PR is a possible solution for issue #99

benjaminrigaud:
LGTM

SongoQ:
ping @gingerwizard

gingerwizard:
Sorry been away @SongoQ @marji.

The build fails - but because of the changes. I'll merge now and resolve the templaye issue.

PR #113, opened 2016-06-18
Improve idempotency for plugins management

barryib:
Don't install plugin when it exists and without asking it explicitly by setting `es_plugins_reinstall: true`. This make the plugin management idempotent.

PS: The filter is taken from the ansible's [elasticsearch_plugin](https://github.com/barryib/ansible-modules-extras/blob/aaa34b03c11d2b53e6a7dfe209423b9da312d250/packaging/elasticsearch_plugin.py)

### DIFF
--- a/filter_plugins/custom.py
+++ b/filter_plugins/custom.py
@@ -19,8 +19,28 @@ def append_to_list(values=[], suffix=''):
 def array_to_str(values=[],separator=','):
     return separator.join(values)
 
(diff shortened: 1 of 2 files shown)


barryib:
closing this. I think that the proper way is to use the [elasticsearch_plugin](http://docs.ansible.com/ansible/elasticsearch_plugin_module.html) module which come with ansible 2.0

PR #115, opened 2016-07-08
Implemntation of elasticsearch_plugin module to handle plugin install…

sirkjohannsen:
…ation

The old way would break on the 2nd run when plugins where already installed

### DIFF
--- a/library/elasticsearch_plugin.py
+++ b/library/elasticsearch_plugin.py
@@ -1,0 +1,116 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2016, simplesurance GmbH, Sirk Johannsen <sirk.hjohannsen@simplesurance.de>
+#
+# This file extends Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


sirkjohannsen:
@elasticsearch-release  jenkis, test it

barryib:
@sirkjohannsen why don't you use the official [elasticsearch_plugin](http://docs.ansible.com/ansible/elasticsearch_plugin_module.html) module for ansible? It's available since the 2.0 version.

gingerwizard:
@barryib @sirkjohannsen this seems to just use the module itself by explictly importing the python lib.  I would rather we just use the module and test it.

barryib:
@gingerwizard I actually use the module from ansible-modules-extras in production. Furthermore, it now supports proxy with a pull request I made.

gingerwizard:
@barryib agreed closing.  Feel free to make a pull request using it. +1 to use.

sirkjohannsen:
@barryib  true. thanks for pointing me to it

barryib:
Hello @sirkjohannsen @gingerwizard,

I'm actually using a modified version of the module. The official have some bugs:
- Avoid using parameters when they are not needed ([#1785](https://github.com/ansible/ansible-modules-extras/issues/1785))
- Fix bug when using proxy ([#2603](https://github.com/ansible/ansible-modules-extras/issues/2603))

So I made the PR [#2604](https://github.com/ansible/ansible-modules-extras/pull/2604) to solve those bugs. But I think it'll only available in the next release of ansible 2.1.1.

I was wondering if it's better to wait that version before using the **elasticsearch_plugin** module.

What do you think?

gingerwizard:
@barryib agreed.  We also need to add tests.  Please can we start a new issue to track moving to either this plugin or a better implementation.  I would like to use for https://github.com/elastic/ansible-elasticsearch/issues/124 which i have almost finished the first implementation of.  Currently it uses the old plugin installation method.

PR #127, opened 2016-07-24
Fail template installation task if HTTP response isn't 200

aosagie:


barryib:
@aosagie you should signed the CLA before doing merge request.

tasks/elasticsearch-templates.yml
```diff
@@ -29,5 +29,7 @@
   register: resultstemplate
 
 - name: Install template(s)
-  command: "curl -sL -XPUT http://localhost:{{http_port}}/_template/{{item}} -d @/etc/elasticsearch/templates/{{item}}.json"
+  command: "curl -sL -XPUT --output /dev/null --write-out %{http_code} http://localhost:{{http_port}}/_template/{{item}} -d @/etc/elasticsearch/templates/{{item}}.json"
```
barryib: I think it would better to the [uri module](http://docs.ansible.com/ansible/uri_module.html) to interact with the API.

karmi:
Hi @aosagie, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/127.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

aosagie:
@karmi Thanks for clearing that up. I was getting confused

tasks/elasticsearch-templates.yml
aosagie: @barryib Agreed, I was just aiming for the minimal change necessary because I wasn't sure of the full context for this decision (curl vs. uri module). Anyway, I see you raised an issue to switch so I'll just close this PR.

gingerwizard: @aosagie thanks for pull request. I'm doing a larger pull which moves templates to a handler. I'll use the uri module as suggested by @barryib. Thanks for the reminder.

PR #129, opened 2016-07-24
Xpack Support (inc. Shield) and Various Improvements

gingerwizard:
This isn't quite complete.  I think it needs idempotent tests in addition to a few more generic tests.  I will document once feedback has been collected as its a significant change.  In summary:

The changes focus on making the role more focused on supporting xpack moving forward.  We therefore no longer treat xpack features the same as plugins.  They will be a single plugin in 5.0 and require distinct configuration. They also now move lock-step with the ES version. This pull request adds xpack feature support inc. shield for versions > 2.0 only.  Other non-xpack plugins will be supported using the same approach as before.  Users wishing to install watcher or marvel on versions < 2.0 should use the plugin arch.

Currently we rely on 2.0.2 ansible due to issue in 2.1 - [handlers cannot use includes](https://github.com/ansible/ansible/issues/15915).  Likely we will release with a dependency on > 2.1.1.

Shield support only supports native and file realm currently.  role-mappings and SSL support should be easy to add.

As suggested by @barryib users and roles are declared as variables.  I structured as required by the realm file - this conveniently serialises to the native request format as well.  Example [here](https://github.com/gingerwizard/ansible-elasticsearch/blob/master/defaults/main.yml)

**Overview of other key changes:**

1. All parameters now set in elasticsearch-parameters where possible allowing us to drop out early. Small refactor.
2. Separate plugin directory support  has been removed - this breaks shield and means we are now consistent with the puppet module.  Instances on the same machine will share a plugin directory like they do an es_home.  Consulting with elastic devs this seems acceptable.
3. Xpack enabled through the switch es_enable_xpack.  This installs license for you.
4. Shield features listed through a [es_xpack_features variable.](https://github.com/gingerwizard/ansible-elasticsearch/blob/master/test/integration/xpack.yml#L11)
5. If enabling shield, we now require an admin user and password to proceed (es_api_basic_auth_username and es_api_basic_auth_password).  This is required for all http calls e.g. loading templates or native users.  If you don't provide we error.  If deploying shield this should be created with the file realm so a user exists on startup.
6. Gem changes to latest versions for test infrastructure. 
7. Users and roles should be declared under a realm - example [here](https://github.com/gingerwizard/ansible-elasticsearch/blob/master/test/integration/xpack.yml#L16).  For roles this just determines how they are added.
8. The role will remove users and roles not declared in a native or file block.  
9. xpack feature removal is supported - just don't list and any features are removed.
10. The role uses es_api_host and es_api_port to control where commands are issued over http e.g. on template load, rather than assuming localhost and 9200.  These default to localhost and 9200 but should be [set](https://github.com/gingerwizard/ansible-elasticsearch/blob/master/defaults/main.yml#L27) if the host and port changes.
11. Es version is now 2.3.4 by default.
12. We use the uri module. httplib2 installed with pip as required. Likely will be removed if we require > 2.1 ansible.
13. Templates are now loaded on the end as a handler.  This is required for subtle reasons.  If the user enables shield, and the node is running, the templates must only be loaded after a restart.  To minimise restarts we do as a handler.  Likewise for native users.
14. Shield config goes in es_config variable.

@elasticdog @dliappis @tylerjl  @barryib @jakommo @strootman


gingerwizard:
Some build failures i will fix and update pull.  Shouldn't impact overall approach.

gingerwizard:
Ok the only test failure is infra based
http://build-eu-00.elastic.co/job/ansible_elasticsearch_acceptance_pr/178/OS=centos-7,es_version=1x,label=metal-pool,suite=config/console

@dliappis can this be resolved at a test image level?

tasks/elasticsearch-Debian.yml
```diff
@@ -26,3 +26,7 @@
   apt: deb=/tmp/elasticsearch-{{ es_version }}.deb
   when: not es_use_repository
   register: elasticsearch_install_from_package
+
+# ansible uri module requires httplib2
+- name: pip httplib2
+  pip: name=httplib2 extra_args="--user"
```
barryib: I don't that everybody has pip installed in their production environment. In debian there is [python-httplib2](https://packages.debian.org/fr/sid/python-httplib2).

For info, in Ansible 2.1 (taken from the [uri module doc](http://docs.ansible.com/ansible/uri_module.html#notes)): 

> The dependency on httplib2 was removed in Ansible 2.1

tasks/elasticsearch-RedHat.yml
```diff
@@ -20,3 +20,7 @@
   yum: name={% if es_custom_package_url is defined %}{{ es_custom_package_url }}{% else %}{{ es_package_url }}-{{ es_version }}.noarch.rpm{% endif %} state=present
   when: not es_use_repository
   register: elasticsearch_install_from_package
+
+# ansible uri module requires httplib2
+- name: pip httplib2
+  pip: name=httplib2 extra_args="--user"
```
barryib: Same as Debian, you can use **python-httplib2** package.

tasks/elasticsearch-Debian.yml
gingerwizard: hopefully we can remove this all together and only support 2.1+
if not, i'll swap in python-httplib2 thanks

handlers/elasticsearch-templates.yml
```diff
@@ -0,0 +1,34 @@
+---
+
+- name: Ensure elasticsearch is started
+  service: name={{instance_init_script | basename}} state=started enabled=yes
+
+- name: Wait for elasticsearch to startup
+  wait_for: port={{es_api_port}} delay=10
+
+- name: Get template files
+  shell: find . -maxdepth 1 -type f | sed "s#\./##" | sed "s/.json//" chdir=/etc/elasticsearch/templates
```
barryib: You should use the [find module](http://docs.ansible.com/ansible/find_module.html) available in ansible 2. 

There is a ``patterns`` option if you want to filter out *.json file.

gingerwizard: Agreed legacy from when this was first created. Lets update. thanks

handlers/elasticsearch-templates.yml
```diff
@@ -0,0 +1,34 @@
...
+  register: resultstemplate
+
+- name: Install templates without auth
+  uri:
+    url: http://{{es_api_host}}:{{es_api_port}}/_template/{{item}}
+    method: PUT
+    status_code: 200
+    body_format: json
+    body: "{{ lookup('file', '/etc/elasticsearch/templates/'+item+'.json') }}"
```
barryib: why you did you remove **.json** from the filename if you have to add it again? You can use ``splitext`` filter

gingerwizard: Just legacy from the previous implementation- i replaced the curl only but its a good opportunity to clean  it all up.

handlers/shield/elasticsearch-shield-native.yml
```diff
@@ -0,0 +1,121 @@
...
+#List current users
+- name: List Native Users
+  uri:
+    url: http://{{es_api_host}}:{{es_api_port}}/_shield/user
+    method: GET
+    user: "{{es_api_basic_auth_username}}"
+    password: "{{es_api_basic_auth_password}}"
+    force_basic_auth: yes
+    status_code: 200
+  register: user_list_response
```
barryib: You should consider setting the ``body_format`` to ``json`` because you're dealing with json data. This will also load automatically the JSON data into a json dict.

handlers/elasticsearch-templates.yml
```diff
@@ -0,0 +1,34 @@
...
+  when: not es_enable_xpack or not es_xpack_features is defined or not '"shield" in es_xpack_features'
+  with_items: "{{ resultstemplate.stdout_lines }}"
+
+- name: Install templates with auth
+  uri:
+    url: http://{{es_api_host}}:{{es_api_port}}/_template/{{item}}
+    method: PUT
+    status_code: 200
+    user: "{{es_api_basic_auth_username}}"
+    password: "{{es_api_basic_auth_password}}"
```
barryib: Is the password printed out when the failed or succeed? If yes, shouldn't we set ``no_log: True``?

tasks/elasticsearch-Debian.yml
dliappis: @gingerwizard I have been thoroughly investigating [the above mentioned failure](http://build-eu-00.elastic.co/job/ansible_elasticsearch_acceptance_pr/178/OS=centos-7,es_version=1x,label=metal-pool,suite=config/console) and as always with such things haven't been able to reproduce it locally.

However, the core issue is that the (fedora maintained) EPEL repos are notoriously busy and frequently unavailable. When the role tries to install elasticsearch, in CentOS, yum checks the repo metadata for all defined repos and fails for EPEL. Lots of other people are being affected by it too, e.g. [in this issue](https://github.com/geerlingguy/ansible-role-repo-epel/issues/8) with Jeff Geerling's EPEL galaxy role. Our CI docker images use the EPEL repo as part of their build process as this is required to install ansible among other things.

There are two things we can do:

1. Remove the EPEL and other unnecessary repos for CentOS, referenced by any `yum` operation as a pre-test kitchen step. I will be pushing a PR for that momentarily.

2. As there is no 100% network guarantee for network or repo availability for anything, it would make sense to add a retry mechanism for all `yum install` tasks. See [this example](https://github.com/geerlingguy/ansible-role-repo-epel/issues/8#issuecomment-124218341)
This will shield the role from failing (immediately) when yum faces transient network issues on ec2 and elsewhere.

README.md
```diff
@@ -251,7 +251,6 @@ controlled by the following parameters:
 * ```es_data_dirs``` - defaults to "/var/lib/elasticsearch".  This can be a list or comma separated string e.g. ["/opt/elasticsearch/data-1","/opt/elasticsearch/data-2"] or "/opt/elasticsearch/data-1,/opt/elasticsearch/data-2"
 * ```es_log_dir``` - defaults to "/var/log/elasticsearch".
 * ```es_work_dir``` - defaults to "/tmp/elasticsearch".
-* ```es_plugin_dir``` - defaults to "/usr/share/elasticsearch/plugins".
```
barryib: @gingerwizard for information, what motivate the ``es_plugin_dir`` removal?

gingerwizard: @barryib sorry ive been away for a while. Spending the next few days catching up with merges and pull requests.  Basically separate plugin dirs cant be supported per node on a host - it prevents shield from working.  This brings consistency with the puppet module.

gingerwizard:
@barryib https://github.com/rvm/rvm1-ansible/issues/13 prevents using python-httplib2 through yum on centos 7.

gingerwizard:
@barryib @jakommo @strootman merging this into trunk.  We wont release until Ansible version 2.1.2 is released to solve handler [issue](https://github.com/ansible/ansible/issues/15915).  Also pending removal of httplib2.

barryib:
@gingerwizard thanks for your PR and to pointing me to that issue. I ran into it my self. I hope this will be release soon.

PR #136, opened 2016-08-03
Make the Upstart script consistent with the systemd script

bobbrez:


bobbrez:
I signed the CLA, this should be ready to go.

templates/init/debian/elasticsearch.j2
```diff
@@ -103,7 +103,7 @@ fi
 
 # Define other required variables
 PID_FILE="$PID_DIR/$NAME.pid"
-DAEMON=$ES_HOME/bin/elasticsearch
+DAEMON={{es_home}}/bin/elasticsearch
```
strootman: I would have left this part be. The templating at line 52 takes care of this.

bobbrez: With `{{es_home}}` it makes it consistent with how the [existing systemd](https://github.com/elastic/ansible-elasticsearch/blob/master/templates/systemd/elasticsearch.j2#L18) is handling it. I'm fine either way.

gingerwizard:
@bobbrez As far as i can tell none of this is needed as the script reads the default file for its environment settings.  Happy to merge for consistency however althoigh functionally im not sure it brings anything

PR #137, opened 2016-08-10
Use openjdk-8-jre-headless, if ubuntu >= 15.10

jakommo:
Needed because ubuntu 16.04 does not provide openjdk-7 packages anymore, which was the default package.
Now uses openjdk-8 package if ubuntu version >= 15.10


jakommo:
jenkins, test it

jakommo:
@gingerwizard ok to merge? Any comments?

barryib:
@jakommo why not use the `es_java` variable to define this kind of "need"?

PR #138, opened 2016-08-17
Add support for custom Debian repository; allow repository key to be set to false

meersjo:
I have a number of hosts that cannot do direct downloads. This minor change enables simplictic support of custom apt repositories, and also allows the repository signing key url to be set to false for unsigned repositories or for when es_apt_url is a ppa.

I have no rpm distributions to test on, so I have not made the same change to that side of the playbook. It might be worth for someone to do this, though, as it seems like a good idea to move these hardcoded values away from the actual tasks and into the variable defaults.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,8 @@ es_major_version: "2.x"
 es_version: "2.2.0"
 es_version_lock: false
 es_use_repository: true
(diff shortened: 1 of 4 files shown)


gingerwizard:
I agree @meersjo this is a good change. About to do some major merging but this will be merged. Thanks

gingerwizard:
Merging through https://github.com/elastic/ansible-elasticsearch/pull/142

PR #141, opened 2016-08-29
template and init script fixes

sammerry:
1) When binding elasticsearch to aws local dns entry localhost does not respond and breaks template loading.

2) Also, when there's a jvm error no print out is provided while starting using the init script. by removing the `-b` flag jvm errors are returned on a failed start.

Fixes for both use cases.

### DIFF
--- a/tasks/elasticsearch-templates.yml
+++ b/tasks/elasticsearch-templates.yml
@@ -22,12 +22,13 @@
       - always
 
 - name: Wait for elasticsearch to startup
-  wait_for: port={{http_port}} delay=10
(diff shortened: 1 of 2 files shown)


sammerry:
Signed the CLA

gingerwizard:
@sammerry I think this has been resolved in the latest changes.  Templates are now handled through a handler which uses url lib and requires the user to declare an api port.  Please review. The init script is based on the debian package and id be reluctant to change here. Feel free to raise in the main ES repo however.

sammerry:
@gingerwizard this has been brought up in the main repo in at least 2 tickets
https://github.com/elastic/elasticsearch/issues/8796
https://github.com/elastic/elasticsearch/issues/12716

PR #146, opened 2016-09-06
Allow unauthenticated package installs when es_apt_key is false

meersjo:
Additional change related to the custom Debian repositories: if no signing key is defined, assume packages are not signed and set apt::allow_unauthenticated=true; conversely, if the signing key is defined, actively set apt::allow_unauthenticated=false regardless of inherited defaults.

### DIFF
--- a/tasks/elasticsearch-Debian.yml
+++ b/tasks/elasticsearch-Debian.yml
@@ -14,7 +14,7 @@
   when: es_use_repository
 
 - name: Debian - Ensure elasticsearch is installed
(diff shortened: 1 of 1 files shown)


gingerwizard:
@meersjo please sign CLA
@dliappis i'd appreciate your thoughts on this one.

gingerwizard:
@meersjo nudge on the CLA so we can test. 

@jakommo thoughts on this?

meersjo:
@gingerwizard Adobe Sign keeps whining that cookies are disallowed while they're not. Can't be bothered to debug their junk.

Note that original change this relates to was also mine, and merged without a problem.

gingerwizard:
I'm merging this despite the CLA as its minor and was making the change anyway.

Issue #147, opened 2016-09-06
Plugins don't install "Can't open //etc/default/elasticsearch"


bhcopeland:
So, I updated my repo and updated to 2.4.0. I ran the playbook to update:

```
ASK [ansible-elasticsearch : Install elasticsearch plugins] *******************
task path: ansible-playbooks/roles/ansible-elasticsearch/tasks/elasticsearch-plugins.yml:34
basic
skipping: [elk.x.org] => (item={u'version': u'latest', u'plugin': u'mobz/elasticsearch-head'})  => {"changed": false, "item": {"plugin": "mobz/elasticsearch-head", "version": "latest"}, "skip_reason": "Conditional check failed", "skipped": true}
skipping: [elk.x.org] => (item={u'version': u'latest', u'plugin': u'lmenezes/elasticsearch-kopf'})  => {"changed": false, "item": {"plugin": "lmenezes/elasticsearch-kopf", "version": "latest"}, "skip_reason": "Conditional check failed", "skipped": true}
```

Upon looking into it, I went to run /usr/share/elasticsearch/bin/plugin, and got:

```
/usr/share/elasticsearch/bin/plugin: 49: .: Can't open //etc/default/elasticsearch
```

I looked in the playbook and saw:

```
- name: Delete Default Environment File
  file: dest=/etc/default/elasticsearch state=absent
  when: ansible_os_family == 'Debian'
```

barryib:
@bhcopeland can you please run your playbook in verbose and share your error?

Normally, these following env vars are set during plugin management:
```yml
  environment:
    CONF_DIR: "{{ conf_dir }}"
    ES_INCLUDE: "{{ instance_default_file }}"
```
So wont (normally) have this error:
``/usr/share/elasticsearch/bin/plugin: 49: .: Can't open //etc/default/elasticsearch``

If you want to run plugin installation manually, you have to export those environment variables.

For the task you're referring to (**Delete Default Environment File**), its normal. The default ``/etc/default/elasticsearch`` is removed and replaced by a custom one (see [here](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-config.yml#L25))

bhcopeland:
Thanks Barry.

You are right, the modified environment is in place, I was running it manually, because my plugins are not installing with the playbook. When I run the playbook, I get Conditional check failed.

https://gist.github.com/b6ca5e3f3ee93c5a15f211e399018a28

[...]

PR #151, opened 2016-09-09
[WIP] Don't overwrite es_plugins_reinstall but set es_plugins_install instead

barryib:
When using tags or setting explicitly ``es_plugins_reinstall`` to true in vars, this last is overwrite with a ``set_fact``.

With this PR, we don't overwrite ``es_plugins_reinstall``, but we ``set es_plugins_install`` instead to switch the plugin install.

Here is a related issue #147. See [this](https://github.com/elastic/ansible-elasticsearch/issues/147#issuecomment-245434439) comment for details.

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -1,8 +1,8 @@
 ---
 
-# es_plugins_reinstall will be set to true if elasticsearch_install_from_repo.changed or elasticsearch_install_from_package.changed
+# es_plugins_install will be set to true if elasticsearch_install_from_repo.changed or elasticsearch_install_from_package.changed or es_plugins_reinstall is explicitly set to true
 # i.e. we have changed ES version(or we have clean installation of ES), or if no plugins listed. Otherwise it is false and requires explicitly setting.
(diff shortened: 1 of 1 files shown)


bhcopeland:
+1 works here.

Although, when I set `es_plugins_reinstall: false` the plugins are still reinstalled.

gingerwizard:
@barryib if some explicitly sets es_plugins_install: false then this will potentially still result in plugin install.  Maybe make the setting of es_plugins_install conditional on it not being set explictly?

We could of actually done that for es_plguins_reinstall...which might be clearer?

gingerwizard:
jenkins, test it

barryib:
@gingerwizard @bhcopeland I that this is a regression introduced by this commit 53535241370db11c77fac379f547f65fb49b1a13. See this [line](https://github.com/elastic/ansible-elasticsearch/commit/53535241370db11c77fac379f547f65fb49b1a13#diff-3f419eb05dd0502a4e53b27340029a92L5).

In that previous version, we only set ``es_plugins_reinstall`` to true when one of these following statements are true:
- ``elasticsearch_install_from_repo.changed`` - for plugins update
- ``elasticsearch_install_from_package.changed`` - for plugins update
- ``es_plugins`` is none - for plugins removal

In other way, you can explicitly set ``es_plugins_reinstall`` to false to avoid plugins reinstallation if nothing need to be updated.

gingerwizard:
@barryib agreed i think we just revert this...ill close this and do a push now

barryib:
@gingerwizard Thanks.

PR #158, opened 2016-09-19
Activate elasticsearch license and some fixes

barryib:
- We will support xpack soon, so I think we should able to enable xpack by activating the corresponding license by providing `es_xpack_license`
- Don't log user credential on the cli when adding or updating user password
- Fix a typo when removing plugin

@gingerwizard please review and advice :)

### DIFF
--- a/tasks/xpack/elasticsearch-xpack-install.yml
+++ b/tasks/xpack/elasticsearch-xpack-install.yml
@@ -14,7 +14,7 @@
 #Remove Plugin if installed and its not been requested or the ES version has changed
 - name: Remove {{item}} plugin
   command: >
-    {{es_home}}/bin/plugin remove shield
+    {{es_home}}/bin/plugin remove {{item}}
   register: xpack_state
   failed_when: "'ERROR' in xpack_state.stdout"
   changed_when: xpack_state.rc == 0
--- a/handlers/shield/elasticsearch-shield-native.yml
+++ b/handlers/shield/elasticsearch-shield-native.yml
@@ -64,6 +64,7 @@
     password: "{{es_api_basic_auth_password}}"
     force_basic_auth: yes
   when: manage_native_users and es_users.native.keys() > 0
+  no_log: True
   with_dict: "{{es_users.native}}"
 
 #List current roles
--- a/tasks/xpack/elasticsearch-xpack-activation.yml
+++ b/tasks/xpack/elasticsearch-xpack-activation.yml
@@ -1,4 +1,4 @@
-
+---
 - name: Activate ES license (without shield authentication)
   uri:
     method: PUT
@@ -11,7 +11,7 @@
   when: '"shield" not in es_xpack_features
   failed_when: >
     license_activated.status != 200 or
-    license_activated.json.license_status is not defined or 
+    license_activated.json.license_status is not defined or
     license_activated.json.license_status != 'valid'
 
 - name: Activate ES license (with shield authentication)
@@ -29,7 +29,7 @@
   when: '"shield" in es_xpack_features
   failed_when: >
     license_activated.status != 200 or
-    license_activated.json.license_status is not defined or 
+    license_activated.json.license_status is not defined or
     license_activated.json.license_status != 'valid'
 
 - debug:
--- a/tasks/xpack/elasticsearch-xpack-activation.yml
+++ b/tasks/xpack/elasticsearch-xpack-activation.yml
@@ -1,0 +1,36 @@
+
+- name: Activate ES license (without shield authentication)
+  uri:
+    method: PUT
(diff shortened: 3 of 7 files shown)


gingerwizard:
This all looks good @barryib but i think we need to test - i need to think how we can do that without actually inserting a license in the repo.  

Oh and nice catch on the typo!

tasks/xpack/elasticsearch-xpack-install.yml
```diff
@@ -14,7 +14,7 @@
 #Remove Plugin if installed and its not been requested or the ES version has changed
 - name: Remove {{item}} plugin
   command: >
-    {{es_home}}/bin/plugin remove shield
+    {{es_home}}/bin/plugin remove {{item}}
   register: xpack_state
```
gingerwizard: nice catch! ouch

barryib:
@gingerwizard is one of these solutions ok?
- use kitchen-ci's env vars during provisioning. Something like ``<%= ENV['ES_XPACK_LICENSE'] %>``
- include a (maybe a vaulted) trial license

gingerwizard:
I like the env variable option as a test license will require renewals.  I can put a test license for 1.7 and 2.x on the build servers. @barryib 

@jakommo and @jpcarey

gingerwizard:
@tylerjl could we use vault her for license storage?

tylerjl:
@gingerwizard it's certainly a possibility - I'll open up an issue in our internal infra to discuss it, but drawing from an environment variable definitely seems like the most flexible/standard handoff point between secrets and the testing process.

gingerwizard:
K env variable it is. @barryib can you add to the kitchen test please and modify the pull?

barryib:
@gingerwizard ping

I don't have the required environment for tests, but I added a test witch:
- retrieve the license
- check if it's activated
- check if the license uid match with the one provided in env vars ``ES_XPACK_LICENSE_UID``

I'v also moved the activation into a separated file ``xpack/elasticsearch-xpack-activation.yml``

Please review.

gingerwizard:
Thanks @barryib ill put a license on the test servers and change the jenkins job.
I think we might need to test this within the context of multiple nodes.  I need to check but i believe we can insert the same license multiple times i.e. once per node, and ES will accept.

gingerwizard:
@barryib so we have an issue in that i don't want the test to run with a trial license as it could potentially expire resulting in failed tests. This leans me towards using a perpetual license.  However, if this i set in the env variable of the job configuration it will be echoed out to the console - potentially leaking a license to the community (something which would make me very unpopular).

Rather this set this through an env variable can i propose the es_license takes a path to a file.  We could then store a license safely on each slave and simply echo the path to the file.


Btw the test fails locally, 
`
       fatal: [localhost]: FAILED! => {"failed": true, "reason": "Syntax Error while loading YAML.\n\n\nThe error appears to have been in '/tmp/kitchen/elasticsearch/tasks/xpack/elasticsearch-xpack-activation.yml': line 15, column 47, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n    license_activated.json.license_status is not defined or\n    license_activated.json.license_status != 'valid'\n                                              ^ here\n"}
`

barryib:
@gingerwizard I do understand your point for the es_license. But using a simple file means for users that we can't vault the license file (ansible lookup plugin doesn't support vaulted files). Is using a vaulted yml file ok for you?

In the playbook, use something like 
```yml
include_vars: my_vaulted_license_var.yml
```
or though env vars
```yml
include_vars: <%= ENV['ES_XPACK_LICENSE_FILE'] %>
```
I could add this in the tests easily and lets users secure their license with ``ansible-vault``.

For the error, I think it's syntax matter. I'll correct it in my next commit.

gingerwizard:
A vaulted file is nice to have. I'm not sure how sensitive a license is and any tests will need to point to a password file to be automated. Make the pull request and ill run the tests this end and make the infra changes.

barryib:
@gingerwizard is the commit 5da6778 OK for you?

barryib:
@gingerwizard any update on this PR?

Thermi:
ansible-elasticsearch supports xpack now, but not installing the license yet. I think this is important.

gingerwizard:
@barryib this didn't get the attention it deserved. I'll test and try and merge today

gingerwizard:
@barryib see https://github.com/elastic/ansible-elasticsearch/pull/223. I believe this will complete the work and allow us to merge.

barryib:
Great. LGTM

---
Thierno IB. BARRY / IT Ninja

[image: Twitter]  <https://twitter.com/barryib>[image: LinkedIn]
<https://fr.linkedin.com/in/barryib>[image: Github]
<https://github.com/barryib>

2017-01-05 21:20 GMT+06:30 Dale McDiarmid <notifications@github.com>:

PR #172, opened 2016-10-27
remove old http package url

koenpunt:
Followup on #170

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -5,6 +5,7 @@ es_version_lock: false
 es_use_repository: true
 es_apt_key: "https://packages.elasticsearch.org/GPG-KEY-elasticsearch"
 es_apt_url: "deb https://packages.elastic.co/elasticsearch/{{ es_major_version }}/debian stable main"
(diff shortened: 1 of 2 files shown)


jakommo:
Thanks for the PR @koenpunt , unfortunately it doesn't work like that.
Results in:
```
 [WARNING]: While constructing a mapping from ..../ansible-elasticsearch/tasks/elasticsearch-Debian.yml, line 17, column 3, found a duplicate dict key (apt_repository).  Using last defined value only.
```
And then both repos end up in the file again. 

It would work with:

```
- name: Debian - Add elasticsearch repository
  apt_repository: repo={{ item.repo }} state={{ item.state}}
  with_items:
    - { repo: "{{ es_apt_url_old }}", state: "absent" }
    - { repo: "{{ es_apt_url }}", state: "present" }
  when: es_use_repository
```

But I think we should make use of the `filename=` setting of `apt_repository` as it currently generates `packages_elastic_co_elasticsearch_2_x_debian.list` and with 5.0 just released we will face a similar issue soon, if we do not use a filename that is not bound to a version.

koenpunt:
Ah, of course. Did not test it 💩

PR #175, opened 2016-10-28
Use network.host and http.port if available for templates and scripts

jakommo:
Was causing issues if not bound to localhost and using none default port, as es_api_host/port default to localhost/9200.

@gingerwizard if you got a minute, a review would be nice, as you initially wrote this an I might have missed something.


jakommo:
Jenkins, test it!

jakommo:
Jenkins, test it!

PR #177, opened 2016-11-02
Remove debug line

veger:
fbfbb66d564e5c1696e2803f2310d5d7d866b6e4 added a debug line at the bottom of the custom.py filter which prints a lot of `basic`.

When this line is removed, the output is normal again.


jakommo:
Jenkins, test it!

PR #178, opened 2016-11-02
Add binary prefix for version 5.x

veger:
The `plugin` executable for Elasticsearch version 5.0.0, is renamed `elasticsearch-plugin`.

This PR uses the new executable name when `es_version` indicates that version 5 (or higher) is used.

Some other minor plugin-related changes:

* Removed `sed -n '1!p' | cut -d '-' -f2-` as the executable just returns a list of installed plugins (or is this also new in version 5.x?)
* Added task name for 'installed plugins check'
* Result of 'remove plugins' task is stored in `plugin_removed`

Note that I have updated the xpack/* files as well, but this is untested. Since it also contains some shell filters (`tail -n +2`), I would assume it does not work with version 5 as well...

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -13,7 +13,7 @@
 - set_fact: es_binary_prefix=""
 
 - set_fact: es_binary_prefix="elasticsearch-"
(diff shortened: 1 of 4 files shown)


veger:
Added some fixes which prevented me to install ES version 2.3

PR #179, opened 2016-11-02
Prevent Ansible errors when shield is not installed/disabled

veger:
When shield is not installed/disabled, i.e. the _elasticsearch-shield-file.yml_ os not included (due to the `when`-statement), the `with_items` and `with_dict` statements are still evaluated (so Ansible is able to check the `when` for each item in the list/dict).

Unfortunately, the `users_to_remove`, `users_to_add`, and `es_users` variables are not present, so the evaluation of the `with_items` and `with_dict` lists fail.

This PR adds defaults to fix these error situations.

When shield is installed/active the variables are present and the defaults are not used.

Note that I am using Ansible 2.2 (although this issue is also reproducable with ansible 2.1). Maybe Ansible behavior with respect to `include` changed?

### DIFF
--- a/tasks/xpack/shield/elasticsearch-shield-file.yml
+++ b/tasks/xpack/shield/elasticsearch-shield-file.yml
@@ -16,7 +16,7 @@
   command: >
     {{es_home}}/bin/shield/esusers userdel {{item}}
   when: manage_file_users and (users_to_remove | length > 0)
-  with_items: "{{users_to_remove}}"
+  with_items: "{{users_to_remove | default([])}}"
   environment:
     CONF_DIR: "{{ conf_dir }}"
     ES_HOME: "{{es_home}}"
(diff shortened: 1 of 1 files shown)


jakommo:
Jenkins, test it!

veger:
So 3 tests (environments) failed...

For example Ubuntu 14.04 (using Ubuntu 16.04 myself, hence I started to look at these results):
* A gem is missing?
* Error occurred while cleaning up the tests (something to do with aufs)

I am unsure what to do/how to fix

jakommo:
Jenkins, test it!

jakommo:
@veger looks like it might not be related to the PR, but to an issue with the test system. 
Let's try again. Oh and thanks for the PR :+1:

veger:
You're right, the test now finishes successfully.

PR #184, opened 2016-11-03
Store Debian and RedHat installation results in a separate variable

veger:
I want to check if a Debian package install resulted in a _change_. But the (skipped) RedHat install registers the result (_skipped_). Thereby overriding the actual Debian package changes.

This PR stores the results of both tasks in different variables, so both results are available.

:warning: This PR (merge) conflicts with #181, either PR needs to be updated (when the other is accepted)

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,7 +1,7 @@
 
 - name: restart elasticsearch
   service: name={{instance_init_script | basename}} state=restarted enabled=yes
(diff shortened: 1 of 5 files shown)


ushuz:
The other registered variable `elasticsearch_install_from_package` may also be affected.

gingerwizard:
@veger im going to try and test this locally as our test infrastructure needs some attention im working on in parallel.  If it succeeds ill merge through a PR.

veger:
I rebased this PR/commit for current master, in order to help out smoothen the process

gingerwizard:
Merged through #220 thanks @veger

PR #194, opened 2016-11-16
Install python-pip package

smcavoy:
Installs pip. Might need EPEL added to requirements


gingerwizard:
@smcavoy we removed the need for python-httplib2 as ansible v 2.12, which we require, no longer depends.

PR #203, opened 2016-12-05
Install EPEL and python-pip on Red Hat Family OS

luisamador:
When on a Red Hat family OS, there's a problem with this task:

# ansible uri module requires python-httplib2
- name: python-httplib2
pip: name=httplib2

It fails if pip is not installed on the system before running this ansible-elasticsearch role.

Solution:

1) Intall EPEL (repository that contains pip)
2) Install python-pip
3) Install httplib2

### DIFF
--- a/tasks/elasticsearch-RedHat.yml
+++ b/tasks/elasticsearch-RedHat.yml
@@ -24,6 +24,13 @@
   when: not es_use_repository
   register: elasticsearch_install_from_package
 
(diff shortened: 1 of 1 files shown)


gingerwizard:
@luisamador in future you will need to sign the CLA

With reference to this particular pull request we just removed the python-httplib2 as its no longer required in ansible > 2.12.

PR #182, opened 2016-11-03
Add support for Elasticsearch 5.x

veger:
In order to be able to install a Elasticsearch cluster using version 5.0.0, I had to add the following v5.x support (next to the other PRs I submitted today and yesterday):

* **Use log4j2.properties instead of logging.yml** - https://www.elastic.co/guide/en/elasticsearch/reference/5.0/settings.html#logging
* **Add jvm.options** - https://www.elastic.co/guide/en/elasticsearch/reference/master/setting-system-settings.html#es-java-opts
* **Use bootstrap.memory_lock instead of bootstrap.mlockall** - https://www.elastic.co/guide/en/elasticsearch/reference/master/setup-configuration-memory.html#mlockall
* **Get rid of work directory** - https://github.com/elastic/elasticsearch/pull/10672
* **Executable does not accept `-D` anymore, replace with `-E` and `-p`** - `--help` of executable
* **Executable does not accept `--` anymore, replace with `-E`** - https://www.elastic.co/guide/en/elasticsearch/reference/5.0/breaking_50_settings_changes.html#_removed_using_double_dashes_to_configure_elasticsearch
* **Get rid of 5.x default (file)s** - Similar to other package defaults

:warning: I am absolutely no Elasticsearch expert (or a experiences user), so I added my references to the list of changes to show why I made the modifications

Additional remarks:

* I tried to keep ES < 5.x compatible, hence the version checks. If support for < 5.x is not required, I can remove the checks
* I do not know about (the/Ansible) role tests, so they probably break... I would appreciate it, if someone is willing to help, or point me in the right direction!

PR #207, opened 2016-12-05
Add binary prefix for version 5.x

veger:
As requested in #182 a PR for the 5.x branch.

Copied from #178:

The `plugin` executable for Elasticsearch version 5.0.0, is renamed `elasticsearch-plugin`.

This PR uses the new executable name when `es_version` indicates that version 5 (or higher) is used.

Some other minor plugin-related changes:

* Removed `sed -n '1!p' | cut -d '-' -f2-` as the executable just returns a list of installed plugins (or is this also new in version 5.x?)
* Added task name for 'installed plugins check'
* Result of 'remove plugins' task is stored in `plugin_removed`

Note that I have updated the xpack/* files as well, but this is untested. Since it also contains some shell filters (`tail -n +2`), I would assume it does not work with version 5 as well...

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -13,7 +13,7 @@
 - set_fact: es_binary_prefix=""
 
 - set_fact: es_binary_prefix="elasticsearch-"
-  when: es_version | version_compare('2.0', '>')
+  when: es_version | version_compare('5.0', '>=')
 
 #List currently installed plugins - ignore xpack if > v 2.0
 - name: Check installed elasticsearch plugins
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -10,8 +10,14 @@
 - set_fact: list_command="--list"
   when: es_version | version_compare('2.0', '<')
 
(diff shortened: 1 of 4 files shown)


jakommo:
Jenkins, test it!

veger:
Fixed situation where `es_plugins` is undefined, resulting in `es_binary_prefix` being undefiend as well. Now `es_binary_prefix` defined no matter what.

jakommo:
Jenkins, test it!

veger:
While setting up the (failing) test, the _rspec-retry_ ruby gem was not found, is it possible that this influences the CentOS 7 standard-1x test? (as this is now the only test that fails)

jakommo:
Looks like the real problem was that the node did not start:

```
Node listening
listening in port 9200 (FAILED - 1)

version check
should be reported as version 1.7.3 (FAILED - 2)
```

Not sure if this is caused by the rspec-retry ruby gem.
let's run the test again .

jakommo:
Jenkins, test it!

veger:
Looks like it was an hiccup in the test scripts

jakommo:
Merged, thanks again @veger

PR #209, opened 2016-12-06
Only use es_api_host in API calls since es_config['network.host'] is not a reliable source

pixelrebel:
Sticking with the `es_api_host` as the only variable to use for the host in API calls, etc.  `es_api_host ` defaults to `localhost` which is a good default since many elasticsearch configs set `_local_` in `network.host`.

### DIFF
--- a/handlers/elasticsearch-templates.yml
+++ b/handlers/elasticsearch-templates.yml
@@ -3,8 +3,41 @@
 - name: Ensure elasticsearch is started
   service: name={{instance_init_script | basename}} state=started enabled=yes
 
+- set_fact:
+    es_config_host: "{{ item }}"
+  with_items: "{{ es_config['network.host'] }}"
+  when: 
+    - es_config['network.host'] is defined and es_config['network.host'] is not string
+    - not item | match('^_.*_$')
+    - es_config_host is not defined
+  
+- set_fact:
+    es_config_host: "{{ es_config['network.host'] }}"
+  when: 
+    - es_config['network.host'] is defined and es_config['network.host'] is string
+    - not es_config['network.host'] | match('^_.*_$')
+
+- set_fact:
+    es_config_port: "{{ es_config['network.port'] }}"
+  when: 
+    - es_config['network.port'] is defined and es_config['network.port'] is number
+
+- set_fact:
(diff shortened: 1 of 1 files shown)


pixelrebel:
Signed the CLA

jakommo:
Hi @pixelrebel this was the default, but it was causing issues when not bound to localhost and or using a custom port. That's why I changed this in https://github.com/elastic/ansible-elasticsearch/commit/5212c6eb4c84b04b36f5850740809d26d6658c16#diff-7f25ff5bf33de039104c5c59d1709488

Being  a distributed system, the use of `_local_` would only make sense if it's a single node install, but the main goal for this playbook is to make it easy to manage a cluster with several nodes.

So I'm not going to revert it, but I'm open to changes that make it work in both cases. 
One thing that comes to mind is having a check if `network.host` is an IP address.

pixelrebel:
>One thing that comes to mind is having a check if network.host is an IP address.

Perhaps checking for the special `_underscore_` notation?  This way, people can still use DNS if they prefer.  Since `network.host` one of those hybrid string/list types.  The logic for properly testing the value(s) could get pretty ugly in ansible. Perhaps it's a standalone task, registering a new variable for use in the URI task?  

Also, if the plan is derive the host name from the ES configs, then is there any reason for setting `es_api_host`?  It could just cause confusion.  Maybe just a fallback to `localhost:9200` is sufficient?

pixelrebel:
@jakommo this is working pretty well for me, what do you think?

jakommo:
Jenkins, test it!

jakommo:
Jenkins, test it!

jakommo:
A lot of the tests failed, seems to be caused by the template check. I need to look at this in more detail

gingerwizard:
taking this on. @pixelrebel the use of es_api_host is a good fix. However, i don't see why you dont use the original means of passing the files through a file path - why do we need to place the file contents into es_template_files_contents? It seems unneccessary and is causing the tests to fail. I will pull your changes and fix before pushing a modified.

gingerwizard:
Ok Re thinking this i think the original approach is the cleaner approach - to rely on es_api_host and es_api_port.  The original issue was the use of these was poorly documented.  We'll close that gap and i'll push a fix.

PR #210, opened 2016-12-07
added ES_JVM_OPTIONS to systemd template

GenmillsKM:
You need to set ES_JVM_OPTIONS to point to the jvm.options file since default is /etc/elasticsearch/.
For installations with multiple roles on single nodes, the jvm.options can go in each roles config directory.
ex. /etc/elasticsearch/master/jvm.options

source: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_packaging.html#_jvm_options

### DIFF
--- a/templates/systemd/elasticsearch.j2
+++ b/templates/systemd/elasticsearch.j2
@@ -10,6 +10,7 @@ Environment=CONF_DIR={{conf_dir}}
 Environment=DATA_DIR={{ data_dirs | array_to_str }}
 Environment=LOG_DIR={{log_dir}}
 Environment=PID_DIR={{pid_dir}}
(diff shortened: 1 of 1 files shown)


karmi:
Hi @GenmillsKM, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/210.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jakommo:
Jenkins, test it!

gingerwizard:
We don't currently have 5.x tests. I'm going to merge this and work on the tests + infra separately.

PR #214, opened 2016-12-27
set ES_JVM_OPTIONS so that the startup script will read the jvm.optio…

alessandroniciforo:
The startup `script /usr/share/elasticsearch/bin/elasticsearch` wasn't able to start elasticsearch with the options defined in jvm.options because the env variable `ES_JVM_OPTIONS` was not defined.
I defined the variable in the config file inside `/etc/sysconfig` and exported it in the `init.d` script

### DIFF
--- a/templates/elasticsearch.j2
+++ b/templates/elasticsearch.j2
@@ -8,6 +8,9 @@ ES_HOME={{es_home}}
 # Elasticsearch configuration directory
 CONF_DIR={{conf_dir}}
 
+# Elasticsearch jvm options file
+ES_JVM_OPTIONS={{conf_dir}}/jvm.options
+
 # Elasticsearch data directory
 DATA_DIR={{ data_dirs | array_to_str }}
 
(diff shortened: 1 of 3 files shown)


gingerwizard:
Including this in a larger pull request that aims to fix all tests on master (now 5.x). Thanks @ alessandroniciforo for the fix!

gingerwizard:
@alessandroniciforo we're going to fix this as part of https://github.com/elastic/ansible-elasticsearch/pull/234.
In summary we'll add the export of the ES_JVM_OPTIONS to the default file and update the start scripts (+ systemd) to use. Nice catch btw. Thanks

@jakommo we need to also test jvm options are applied.

gingerwizard:
resolved through https://github.com/elastic/ansible-elasticsearch/pull/234

karmi:
Hi @alessandroniciforo, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/214.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

alessandroniciforo:
@gingerwizard looks great!

@karmi thanks for letting me know. I just added the other e-mail.

PR #216, opened 2016-12-30
Use elasticsearch_plugin instead of the current shell/command modules

dmizelle:
The current way that elasticsearch plugins are installed caused our internal tests to fail idempotency when using this role.

I've migrated to using elasticsearch_plugin, which locks anyone down to using only Ansible >2.x, but I figure its 'modern,' but may break backwards-compatibility for those using an old ansible install.

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -1,47 +1,15 @@
 ---
-
-# es_plugins_reinstall will be set to true if elasticsearch_install_from_repo.changed or elasticsearch_install_from_package.changed
-# i.e. we have changed ES version(or we have clean installation of ES), or if no plugins listed. Otherwise it is false and requires explicitly setting.
-- set_fact: es_plugins_reinstall=true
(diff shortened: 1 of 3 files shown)


gingerwizard:
thanks @dmizelle we've been meaning to move to this plugin for a while so thanks for taking the initiative.  The previous provided support for removing all plugins and reinstalling through es_reinstall_plugins.  It also did this when moving between ES versions (necessary unfortunately).

Does the plugin address this internally? If not, i think we'll need to enchance this PR.

kireevco:
@dmizelle @gingerwizard Any chance to merge it?

gingerwizard:
@kireevco  see the above concerns. We need to add appropriate tests for plugin reinstall and ability to diff changes. I'll investigate this week and look to either merge or enhance.

gingerwizard:
Discussion https://github.com/elastic/ansible-elasticsearch/issues/125

gingerwizard:
@dmizelle licensing prevents us from merging.

PR #222, opened 2017-01-04
Fix issues with systemd

gingerwizard:
Merges fixes from @vegar as well as issue preventing plugin install on redhat


vegar:
I guess @veger deserves more creds then me on this one ;-)

gingerwizard:
oops thanks!

PR #224, opened 2017-01-05
Version change and X-Pack Docs

gingerwizard:
Moves default version to 2.4.3 and doc updates re x-pack.


gingerwizard:
Xpack failures due to license. Updated license.

jenkins, test it

PR #231, opened 2017-01-08
quote interpolated variables in command

supercoffee:
Ansible 2.2.0 will barf when it hits this task.  It complains about the syntax of unquoted variable. 

The error

```
TASK [elasticsearch : include] *************************************************
fatal: [127.0.0.1]: FAILED! => {"failed": true, "reason": "Syntax Error while loading YAML.\n\n\nThe error appears to have been in '/vagrant/ansible/roles/elasticsearch/tasks/xpack/elasticsearch-xpack-install.yml': line 16, column 23, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n- name: Remove {{item}} plugin\n  command: {{es_home}}/bin/plugin remove {{item}}\n                      ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}
fatal: [127.0.0.1]: FAILED! => {"failed": true, "reason": "Syntax Error while loading YAML.\n\n\nThe error appears to have been in '/vagrant/ansible/roles/elasticsearch/tasks/xpack/elasticsearch-xpack-install.yml': line 16, column 23, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n- name: Remove {{item}} plugin\n  command: {{es_home}}/bin/plugin remove {{item}}\n                      ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}
[...]

PR #212, opened 2016-12-08
Updated yum repository URL for 5.x

JanneSalo:
While trying the role on a CentOS 7 box to install Elasticsearch 5.x, I came across this little issue and fixed it. In a nutshell, the `baseurl` (and `gpgkey`) properties in `templates/elasticsearch.repo` don't work with 5.x. This is simply because the repository base URL has changed (see for example <https://www.elastic.co/guide/en/elasticsearch/reference/2.4/setup-repositories.html#_yum_dnf> vs. <https://www.elastic.co/guide/en/elasticsearch/reference/5.0/rpm.html>). This PR changes the URL for 5.x while retaining backwards compatibility - older versions will still get installed from their respective repositories. In my own tests, this seemed to work fine for both 2.x and 5.x.

PR #234, opened 2017-01-16
5.x Compatibility

gingerwizard:
Fixes all issues with 5.x support and provides supporting up to date tests.

Also resolves

https://github.com/elastic/ansible-elasticsearch/pull/231
https://github.com/elastic/ansible-elasticsearch/pull/212


gingerwizard:
@jakommo @barryib  for review

gingerwizard:
Ok some progress - all but 4 pass. Ill take a look at them - they currently seem to issues with ES starting. I can't reproduce the 2 i have tried locally.

tasks/elasticsearch-config.yml
```diff
@@ -86,8 +79,8 @@
 - name: Delete Default Logging File
   file: dest=/etc/elasticsearch/logging.yml state=absent
 
-- name: Delete Default Logging File (5.x)
+- name: Delete Default Logging File
   file: dest=/etc/elasticsearch/log4j2.properties state=absent
 
-- name: Delete Default JVM Options File (5.x)
+- name: Delete Default JVM Options File
```
dekhtyarev: Hello!

I do not see that customized jvm.options was on its default location, as required documentation, #235 and [my commit](https://github.com/dekhtyarev/ansible-elasticsearch/commit/1efdbc8052d786bf92b9a18e07b07cb21f809d8e) .

Could you fix it?

gingerwizard: Well spotted. The start script doesn't seem to take from the config directory but relies on the ES_HOME/conf. I don't see the service not starting though - it may explain the 4 failed tests though.
It seems the new start scripts in the 5.x package actually export this, similar to what @alessandroniciforo  has done here. https://github.com/elastic/ansible-elasticsearch/pull/214/files
Rather than inserting custom logic i think we should update the start scripts and insert the export of ES_JVM_OPTIONS into the defaults file.

gingerwizard: https://github.com/elastic/elasticsearch/pull/21445

dekhtyarev: Thx for information and you work!
Can you update systemd unit also ? (https://github.com/gingerwizard/ansible-elasticsearch/blob/master/templates/systemd/elasticsearch.j2#L7)

gingerwizard:
Pushed some fixes to resolve handler issues. It seems the restart handler was being called the number of times it was notified due to the include.  This seems to be an ansible issue with 2.2.0.  Remaining issues are with the config job. Investigating.

tasks/elasticsearch-config.yml
gingerwizard: the systemd should be fine as it calls the default file

gingerwizard:
After some investigation (big thanks to @dliappis) it seems values in the multi test. It seems like you can't invoke the same role (with different params) in the same play - otherwise the param values become "mixed". It only seems to be an issue if you call flush handlers.  Contemplating next steps:

1. We adjust the documentation and recommend the following structure:

```
---
#Test ability to deploy multiple instances to a machine
- name: Elasticsearch Multi tests
  hosts: localhost
  vars:
    es_scripts: true
    es_templates: true
    es_heap_size: "1g"

  roles:
    - { role: elasticsearch, es_api_port: 9200, es_instance_name: "master", es_data_dirs: ["/opt/elasticsearch/master"], es_config: { discovery.zen.ping.unicast.hosts: "localhost:9300", http.port: 9200, transport.tcp.port: 9300, node.data: false, node.master: true, bootstrap.memory_lock: true } }

#Test ability to deploy multiple instances to a machine
- name: Elasticsearch Multi tests
  hosts: localhost
  vars:
    es_scripts: true
    es_templates: true
    es_heap_size: "1g"

  pre_tasks:
    - debug: msg="Ok now will run the same role on port 9201"

  roles:
    - { role: elasticsearch, es_api_port: 9201, es_instance_name: "node1", es_data_dirs: "/opt/elasticsearch/data-1,/opt/elasticsearch/data-2", es_config: { discovery.zen.ping.unicast.hosts: "localhost:9300", http.port: 9201, transport.tcp.port: 9301, node.data: true, node.master: false } }
#Plugins installed for this test are specified in .kitchen.yml under suite
```

2. Restructure the role to allow for the above limitation. I currently can't find any ansible documentation specifying the correct behaviour.

gingerwizard:
All green merging! 5.x support!

PR #262, opened 2017-03-07
Add check_mode:no to x_pack_installed

meersjo:
Without this, the variable x_pack_installed doesn't get registered in check mode, and consequently playbook execution halts on the task "Remove x-pack plugin" which uses it in the when clause.

### DIFF
--- a/tasks/xpack/elasticsearch-xpack-install.yml
+++ b/tasks/xpack/elasticsearch-xpack-install.yml
@@ -5,6 +5,7 @@
   register: x_pack_installed
   changed_when: False
   failed_when: "'ERROR' in x_pack_installed.stdout"
+  check_mode: no
   ignore_errors: yes
   environment:
     CONF_DIR: "{{ conf_dir }}"


meersjo:
CLA signed.

karmi:
Hi @meersjo, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/262.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

meersjo:
Sigh. So many hoops for one line of code... Email added.

karmi:
Thanks for adding the e-mail, @meersjo! I know it looks weird to require CLA for a simple change like yours, it's just that it's much easier for us to keep track of everything this way. Thanks again!

gingerwizard:
jenkins, test it

gingerwizard:
LGTM. Thanks @meersjo

PR #263, opened 2017-03-13
Add check_mode: no to prevent failing in --check mode

meersjo:
Adding check_mode:no to prevent ansible-playbook failing on missing dict_objects when running with --check.


gingerwizard:
jenkins, test it!

gingerwizard:
testing again as x-pack test failed through no fault of @meersjo 
jenkins, test it!

gingerwizard:
Im confident this doesn't introduce any issues. The single failure is occuring due to ssh issues between the host and container. This seems to be occuring on 2.x runs only.  Will investigate with @dliappis but i think this is good to merge. Thanks @meersjo

PR #265, opened 2017-03-15
Basic FreeBSD support

tbartelmess:
This adds support for FreeBSD to the ansible role. It's pretty much only a few different paths and an rc.d script.

### DIFF
--- a/handlers/elasticsearch-templates.yml
+++ b/handlers/elasticsearch-templates.yml
@@ -7,7 +7,7 @@
   wait_for: host={{es_api_host}} port={{es_api_port}} delay=10
 
 - name: Get template files
-  find: paths="/etc/elasticsearch/templates" patterns="*.json"
+  find: paths="{{es_conf_dir}}/templates" patterns="*.json"
   register: templates
 
 - name: Install templates without auth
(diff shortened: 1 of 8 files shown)


gingerwizard:
Hi @tbartelmess thanks for this. In order to support a new platform we will need kitchen tests. We would in turn need to ensure the jenkins tests are updated.
The addition issue here is we don't support FreeBSD for elasticsearch https://www.elastic.co/support/matrix, so i'm not sure we want to in ansible. cc @dliappis and @jakommo  for opinions.

tbartelmess:
Hi @gingerwizard, thanks for your response. I can see that adding FreeBSD adds a lot of complexity on your side. The biggest issue with the current kitchen tests is that they assume a Linux kernel since they run on Docker.
There is no easy way to run a FreeBSD userland on a Linux Kernel, (the other way around is technically possible when the FreeBSD Linux compat layer is installed).

The main reason I think FreeBSD support in the ansible module is a good idea, is that a lot of other modules depend on Elasticsearch which then often means that FreeBSD users need to do a bunch of minimal modifications to all modules.

The changes required to make FreeBSD work are fairly minimal, (the biggest thing is usally that FreeBSD uses /usr/local/etc instead of /etc for non system settings and a different init script).

PR #267, opened 2017-03-15
Fixes and tests for idempotent behavour

gingerwizard:
Makes minor modifications to fix and test idempotent behaviour of the role.


gingerwizard:
@jakommo @strootman i believe this makes the role fully idempotent.  We now run each test twice and confirm there are no changes.

Issue #213, opened 2016-12-21
Plugin install fails on upgrade when es_plugins and es_xpack_features are used


jakommo:
With the following config

```
    es_plugins:
     - plugin: cloud-aws
    es_enable_xpack: true
    es_xpack_features:
      - marvel-agent
```

 upgrading ES fails with:

```
failed: [10.52.101.221] (item={u'plugin': u'cloud-aws'}) => {
    "changed": false, 
    "cmd": [
        "/usr/share/elasticsearch/bin/plugin", 
        "install", 
        "cloud-aws", 
        "--silent"
    ], 
    "delta": "0:00:06.482577", 
    "end": "2016-12-20 20:12:52.515926", 
    "failed": true, 
    "failed_when_result": true, 
    "invocation": {
        "module_args": {
            "_raw_params": "/usr/share/elasticsearch/bin/plugin install cloud-aws  --silent", 
            "_uses_shell": false, 
            "chdir": null, 
            "creates": null, 
            "executable": null, 
            "removes": null, 
            "warn": true
        }, 
        "module_name": "command"
    }, 
    "item": {
        "plugin": "cloud-aws"
    }, 
    "rc": 64, 
    "start": "2016-12-20 20:12:46.033349", 
    "stderr": "", 
    "stdout": "ERROR: Plugin [marvel-agent] is incompatible with Elasticsearch [2.4.2]. Was designed for version [2.4.1]", 
    "stdout_lines": [
        "ERROR: Plugin [marvel-agent] is incompatible with Elasticsearch [2.4.2]. Was designed for version [2.4.1]"
    ], 
    "warnings": []
}
```

In this example `cloud-aws` fails to install because `marvel-agent` is still installed in the old version.
I created this upstream issue https://github.com/elastic/elasticsearch/issues/22286 for this.

We would need to first remove all plugins, before installing them in the new versions, which is complicated in this case, as plugins are installed by `es_plugins` and `es_xpack_features`.

gingerwizard:
The bigger problem is the list command doesn't list all the plugins installed - it errors on the first one and doesn't even report others.  It makes it difficult to remove all therefore.

gingerwizard:
We might have to simply list the directory contents

gingerwizard:
https://github.com/elastic/ansible-elasticsearch/pull/269

gingerwizard:
https://github.com/elastic/ansible-elasticsearch/pull/269 closes

PR #269, opened 2017-03-16
Support for plugin diffs + improved testing on minor upgrades + fixes

gingerwizard:
1. ES is now 5.2.2
2. Plugin diffing now occurs. If you remove or add plugins to the list, they will be uninstalled/installed respectively.
3. Standard and multi test becomes the only idempotent test. Enhanced to install plugins with minor config changes for idempotent test.
4. Config test confirms config changes if run multiple times. It also tests plugins can be removed and added.
5. Package test confirms we can upgrade between minor versions and tests this with respect to plugins.
6. Xpack tests check plugin upgrade between minor versions
7. We now identify the installed plugins with a ls command. Resolves https://github.com/elastic/ansible-elasticsearch/issues/213.

I would suggest we rename tests at some point to reflect their purpose. Standard should be enhanced to test more idempotent features. Also we can probably condense.

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -5,25 +5,35 @@
 - set_fact: es_plugins_reinstall=true
   when: (((debian_elasticsearch_install_from_repo is defined and debian_elasticsearch_install_from_repo.changed) or (redhat_elasticsearch_install_from_repo is defined and redhat_elasticsearch_install_from_repo.changed)) or (elasticsearch_install_from_package is defined and elasticsearch_install_from_package.changed)) or es_plugins is not defined or es_plugins is none
 
-- set_fact: list_command="list"
+- set_fact: list_command=""
+#If we are reinstalling all plugins, e.g. to a version change, we need to remove all plugins (inc. x-pack) to install any plugins. Otherwise we don't consider x-pack so the role stays idempotent.
+- set_fact: list_command="| grep -vE 'x-pack'"
+  when: not es_plugins_reinstall
 
-#List currently installed plugins - ignore xpack if > v 2.0
+#List currently installed plugins. We have to list the directories as the list commmand fails if the ES version is different than the plugin version.
 - name: Check installed elasticsearch plugins
-  shell: "{{es_home}}/bin/elasticsearch-plugin list | grep -vE 'x-pack'"
+  shell: "ls {{es_home}}/plugins {{list_command}}"
   register: installed_plugins
-  failed_when: "'ERROR' in installed_plugins.stdout"
   changed_when: False
   ignore_errors: yes
   environment:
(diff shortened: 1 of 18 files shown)


gingerwizard:
Ok the config test issues are due to a spelling mistake.  The multi tests depend on jmespath - ill add to the provisonning and repush.

gingerwizard:
@jakommo @strootman @jpcarey mind reviewing.

tasks/elasticsearch-plugins.yml
```diff
@@ -5,25 +5,35 @@
 - set_fact: es_plugins_reinstall=true
   when: (((debian_elasticsearch_install_from_repo is defined and debian_elasticsearch_install_from_repo.changed) or (redhat_elasticsearch_install_from_repo is defined and redhat_elasticsearch_install_from_repo.changed)) or (elasticsearch_install_from_package is defined and elasticsearch_install_from_package.changed)) or es_plugins is not defined or es_plugins is none
 
-- set_fact: list_command="list"
+- set_fact: list_command=""
+#If we are reinstalling all plugins, e.g. to a version change, we need to remove all plugins (inc. x-pack) to install any plugins. Otherwise we don't consider x-pack so the role stays idempotent.
+- set_fact: list_command="| grep -vE 'x-pack'"
```
barryib: @gingerwizard I dont get why you exclude x-pack here. Specially how this makes it idempotent.

gingerwizard: Normally we don't want to consider x-pack in deltas of plugins - as its handled by a seperate task later.  However, if we are doing a complete re-install e.g. version change, we are forced to uninstall it in order to install any other plugins.

gingerwizard: The later x-pack task will inturn install it again. This was a workaround until we refactor the x-pack handling in the role - which id like to do but suspect it will be very disruptive.

barryib: humm! understood. But for the x-pack installation, why can't we just add `x-pack` to the `es_plugins` list? In that way, all plugins are managed the same way and avoid code duplication.

barryib: maybe latter with the x-pack refactoring?

gingerwizard: indeed thats the plan but it meant touching more beyond the remit of this PR. I think its reasonable to want to do this in one place though - although the x-pack stuff does alot more.

gingerwizard:
@jpcarey changes caused a build and 2 failures. These are hard to reproduce and seem to occur intermittently and are associated with templates and auth - this is subject to a rewrite shortly. Lets see if they occur,

jenkins, test it

gingerwizard:
Actually its because when we go up versions we have to uninstall x-pack which might be causing users to be dropped....

PR #271, opened 2017-03-20
Fix adding of plugins when es_plugins_reinstall is false

veger:
* Get (and use) list of plugins that need to be installed

Similar to getting/calculating the list of plugins that need to be removed, except that the `es_plugins` list also can contain additional data.

* Just remove all plugins of the remove list

After determining which plugins need to be removed, the `when` check is not required (anymore)


barryib:
LGTM.

@gingerwizard @jakommo Is this PR going to be merged? Any ETA?

gingerwizard:
I'm returning to working on this role @barryib @veger and will be clearing the backlog over the next few days + moving us compliant with 5.6.

This seems to cover a cornercase i didn't originally anticipate. Ill merge and ensure tests pass locally, fixing other PRs as required.

PR #281, opened 2017-03-29
attempt to restart via systemd if process killed

mwjackson:
this template change configures systemd to attempt restart the process if it dies

as per: https://www.freedesktop.org/software/systemd/man/systemd.service.html

### DIFF
--- a/templates/systemd/elasticsearch.j2
+++ b/templates/systemd/elasticsearch.j2
@@ -26,6 +26,8 @@ ExecStart={{es_home}}/bin/elasticsearch \
                                     -Edefault.path.data=${DATA_DIR} \
                                     -Edefault.path.conf=${CONF_DIR}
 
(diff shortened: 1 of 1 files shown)


mwjackson:
Hi, just wondering what the status of this PR is and if there is anything more I need to do to get it merged?

barryib:
@mwjackson I that this PR should be submit [here](https://github.com/elastic/elasticsearch/blob/master/distribution/src/main/packaging/systemd/elasticsearch.service). This role follow the upstream systemd file.

gingerwizard:
Indeed i agree. We remain consistent with the systemd file distributed by Elasticsearch. Alot of thought goes into these start scripts from the devs with a deep understanding of why this behaviour might not be desirable  - a flapping node could have implications on cluster instability for example.

Issue #259, opened 2017-03-01
Replace -D with -E for proxy specification in x-pack installation


jdlourenco:
The x-pack installation [task ](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/xpack/elasticsearch-xpack-install.yml#L30) is still using the -D option sintax for configuring the proxy that seems to have been deprecated:

`{{es_home}}/bin/elasticsearch-plugin install --silent --batch x-pack {% if es_proxy_host is defined and es_proxy_host != '' %} -Dhttp.proxyHost={{ es_proxy_host }} -Dhttp.proxyPort={{ es_proxy_port }} {% endif %}`

I changed it to the following and it works:

`{{es_home}}/bin/elasticsearch-plugin install --silent --batch x-pack {% if es_proxy_host is defined and es_proxy_host != '' %} -Ehttp.proxyHost={{ es_proxy_host }} -Ehttp.proxyPort={{ es_proxy_port }} {% endif %}`

Best regards,
José Lourenço

jdlourenco:
Is installing plugins behind a proxy even supported in 5.x?

It is not mentioned here https://www.elastic.co/guide/en/x-pack/5.0/installing-xpack.html

strootman:
@jdlourenco Not obvious to find, but check out this link;

https://www.elastic.co/guide/en/elasticsearch/plugins/5.2/_other_command_line_parameters.html


I got there by searching for `plugins` in the [Elasticsearch Reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html), which brings you to the following link;

https://www.elastic.co/guide/en/elasticsearch/reference/5.2/modules-plugins.html

jdlourenco:
Thanks for pointing that out @strootman

So I guess the line should be changed to:

`{% if es_proxy_host is defined and es_proxy_host != '' %} ES_JAVA_OPTS="-Dhttp.proxyHost={{ es_proxy_host }} -Dhttp.proxyPort={{ es_proxy_port }} -Dhttps.proxyHost={{ es_proxy_host }} -Dhttps.proxyPort={{ es_proxy_port }}" {% endif %} {{es_home}}/bin/elasticsearch-plugin install --silent --batch x-pack`

strootman:
@jdlourenco Looks like you're ready for a PR! Then you would get the benefit of all the kitchen.ci tests if you don't want to run them locally.

gingerwizard:
[...]

PR #284, opened 2017-03-31
Change proxy specification for installation of x-pack plugin in ES 5

jdlourenco:
This pull request aims at solving #259 

According to https://www.elastic.co/guide/en/elasticsearch/plugins/5.2/_other_command_line_parameters.html
proxy is now specified using the ES_JAVA_OPT environment variable:

`ES_JAVA_OPTS="-Dhttp.proxyHost=host_name -Dhttp.proxyPort=port_number -Dhttps.proxyHost=host_name -Dhttps.proxyPort=https_port_number" 
`

Also replaced **command** module by **shell** module because **command** wasn't generating the command line correctly.

### DIFF
--- a/tasks/xpack/elasticsearch-xpack-install.yml
+++ b/tasks/xpack/elasticsearch-xpack-install.yml
@@ -27,8 +27,8 @@
 
 #Install plugin if not installed, or the es version has changed (so removed above), and its been requested
 - name: Install x-pack plugin
-  command: >
-    {{es_home}}/bin/elasticsearch-plugin install --silent --batch x-pack {% if es_proxy_host is defined and es_proxy_host != '' %} -Dhttp.proxyHost={{ es_proxy_host }} -Dhttp.proxyPort={{ es_proxy_port }} {% endif %}
+  shell: >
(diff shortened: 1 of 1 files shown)


karmi:
Hi @jdlourenco, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/284.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jdlourenco:
Added.

José Lourenço

2017-03-31 17:02 GMT+01:00 Karel Minarik <notifications@github.com>:

gingerwizard:
@jdlourenco i think we'd rather set this using the environment variables as recommended by ansible [here](http://docs.ansible.com/ansible/latest/playbooks_environment.html) rather than in the shell itself

gingerwizard:
I will make a PR with the above and add a test. Will ensure this is in the PR to credit your welcome contribution. Thanks again for spotting!

barryib:
@gingerwizard don't forget the plugin installation through proxy [here](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-plugins.yml#L50).

gingerwizard:
Closed through https://github.com/elastic/ansible-elasticsearch/pull/357

PR #293, opened 2017-04-20
Upgraded to 5.3.0

igayoso:
Upgraded to last elastic search version (5.3.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ The use of a map ensures the Ansible playbook does not need to be updated to ref
 
 In addition to the es_config map, several other parameters are supported for additional functions e.g. script installation.  These can be found in the role's defaults/main.yml file.
 
-The following illustrates applying configuration parameters to an Elasticsearch instance.  By default, Elasticsearch 5.1.2is installed.
+The following illustrates applying configuration parameters to an Elasticsearch instance.  By default, Elasticsearch 5.3.0 is installed.
 
 ```
 - name: Elasticsearch with custom configuration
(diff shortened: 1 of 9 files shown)


jakommo:
Thanks for contributing @igayoso . Seems like you did not sign the CLA yet? 
I'm a bit concerned to make 5.3.0 the default version, as it is affected by this [path.data bug](https://www.elastic.co/blog/multi-data-path-bug-in-elasticsearch-5-3-0).

I did a quick test, with multiple `path.data` and it seems like we are not affected, since we do not build this as an array:

```
# curl -XGET "http://10.52.101.190:9200/_nodes?pretty&filter_path=nodes.*.settings.path.data"
{
  "nodes" : {
    "tIEECDksRXaCCTD7Pdg1sQ" : {
      "settings" : {
        "path" : {
          "data" : [
            "/mnt/data-1/10.52.101.190-node1",
            "/mnt/data-2/10.52.101.190-node1"
          ]
        }
      }
    }
  }
}
```
and in elasticsearch.yml
```
path.data: /mnt/data-1/10.52.101.190-node1,/mnt/data-2/10.52.101.190-node1
```

However, I'm not sure if there are circumstances where this ended up as an array, so I would feel better to wait for `5.3.1`, which has this fixed and is due to be released very soon.

igayoso:
@jakommo no problem for PR, thanks to you for this repo :) About wait until 5.3.1 I'm totally agree with you. There is an important bug and maybe is not all the stable that we think. I was testing on my environment (not production) and works fine but is not a problem wait until new version. At the end is just change the version number in a few files. I will do when new release is available. Thanks!

jakommo:
Well there we go, 5.3.1 was just released :smile: https://www.elastic.co/blog/elasticsearch-5-3-1-released
Happy to merge the PR if you want to update this one or send in a new one.

igayoso:
@jakommo lol I checked before do PR and it was 5.3.0 yet. I created a new PR with 5.3.1 https://github.com/elastic/ansible-elasticsearch/pull/295

PR #298, opened 2017-05-03
Ability to override path.repo property

remyma:
Add property ```es_repo_dirs``` in defaults so that we can configure backups/restore repositories

### DIFF
--- a/tasks/elasticsearch-parameters.yml
+++ b/tasks/elasticsearch-parameters.yml
@@ -47,3 +47,5 @@
 - set_fact: pid_dir={{ es_pid_dir }}/{{instance_suffix}}
 - set_fact: log_dir={{ es_log_dir }}/{{instance_suffix}}
 - set_fact: data_dirs={{ es_data_dirs | append_to_list('/'+instance_suffix) }}
+- set_fact: repo_dirs={{ es_repo_dirs | append_to_list('/'+instance_suffix) }}
+  when: es_repo_dirs|length > 0
--- a/tasks/elasticsearch-parameters.yml
+++ b/tasks/elasticsearch-parameters.yml
@@ -47,5 +47,3 @@
 - set_fact: pid_dir={{ es_pid_dir }}/{{instance_suffix}}
 - set_fact: log_dir={{ es_log_dir }}/{{instance_suffix}}
 - set_fact: data_dirs={{ es_data_dirs | append_to_list('/'+instance_suffix) }}
-- set_fact: repo_dirs={{ es_repo_dirs | append_to_list('/'+instance_suffix) }}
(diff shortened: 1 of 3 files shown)


tasks/elasticsearch-parameters.yml
```diff
@@ -47,3 +47,5 @@
 - set_fact: pid_dir={{ es_pid_dir }}/{{instance_suffix}}
 - set_fact: log_dir={{ es_log_dir }}/{{instance_suffix}}
 - set_fact: data_dirs={{ es_data_dirs | append_to_list('/'+instance_suffix) }}
+- set_fact: repo_dirs={{ es_repo_dirs | append_to_list('/'+instance_suffix) }}
```
barryib: @remyma Is this a good idea to hase different repo name based on the `inventory_hostname` within the cluster?

I thought that all nodes have to mount the same filesystem to the same location on all master an data node ([more info](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html#_shared_file_system_repository))

remyma: Agree with you. I removed this line

barryib:
@remyma I was wondering if this can't be handled with the existing `es_config` variable like:

```yml
es_config:
  path.repo:
    - /path/to/repo1
    - /path/to/repo2
```

What do you think?

remyma:
I'm not sure this can be done this way. 
Let me clarify this : suppose you have 2 distinct elasticsearch clusters : 
* es_instance_name : monitoring
* es_instance_name : search

Correct me if I'm wrong, but this role allows you to have nodes from different clusters instance on the same host : this is why log and data locations are specific for a cluster instance.
Your backup / restore folder for monitoring cluster should be the same for all nodes of this cluster, but it should be different thant the one used for cluster search. As log and data dirs.  

So finally, I think this can be done by appending instance_suffix (cluster instance name, not inventory hostname) as it is done for data and log folders.

What do you think ?

barryib:
@remyma in that way I think you should have different `es_config` var. Actually one per ES instance.

Taken from the readme:

```yml
- hosts: monitoring
  roles:
    - { role: elasticsearch, es_instance_name: "monitoring", es_heap_size: "1g",
    es_config: {
        cluster.name: "monitoring",
        path.repo: ["/path/to/monitoring/repo"],
        ...
        }
    }

- hosts: search
  roles:
    - { role: elasticsearch, es_instance_name: "search", es_heap_size: "1g",
    es_config: {
        cluster.name: "search",
        path.repo: ["/path/to/search/repo"],
        ...
        }
    }
```

gingerwizard:
I agree with @barryib this is supported and the syntax he highlights is actually the only one we support.

PR #300, opened 2017-05-06
Fixes handler and file issues with ES templates

lobsterdore:
Fixes an issue with handlers in Ansible 2.2.1+
[https://github.com/ansible/ansible/issues/20603](). Moves the
load-templates handler to the top of the file to get around this
issue.

Fixes an issue loading installed templates into ES. The use of
file lookup meant that the template files are checked on the
machine that initiated the Ansible play, this only works if Ansible
has been run locally, breaking for remote machines. Switched to
using the slurp module to grab file contents on the target machine.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,10 +1,4 @@
-- name: reload systemd configuration
-  command: systemctl daemon-reload
-
-# Restart service and ensure it is enabled
-- name: restart elasticsearch
-  service: name={{instance_init_script | basename}} state=restarted enabled=yes
-  when: es_restart_on_change and es_start_service and ((plugin_installed is defined and plugin_installed.changed) or (config_updated is defined and config_updated.changed) or (xpack_state.changed) or (debian_elasticsearch_install_from_repo.changed or redhat_elasticsearch_install_from_repo.changed or elasticsearch_install_from_package.changed))
+---
 
 #Templates are a handler as they need to come after a restart e.g. suppose user removes security on a running node and doesn't
 #specify es_api_basic_auth_username and es_api_basic_auth_password.  The templates will subsequently not be removed if we don't wait for the node to restart.
@@ -13,3 +7,11 @@
 - name: load-templates
   include: ./handlers/elasticsearch-templates.yml
   when: es_templates
+
+- name: reload systemd configuration
+  command: systemctl daemon-reload
+
+# Restart service and ensure it is enabled
+- name: restart elasticsearch
+  service: name={{instance_init_script | basename}} state=restarted enabled=yes
+  when: es_restart_on_change and es_start_service and ((plugin_installed is defined and plugin_installed.changed) or (config_updated is defined and config_updated.changed) or (xpack_state.changed) or (debian_elasticsearch_install_from_repo.changed or redhat_elasticsearch_install_from_repo.changed or elasticsearch_install_from_package.changed))
--- a/handlers/elasticsearch-templates.yml
+++ b/handlers/elasticsearch-templates.yml
@@ -6,29 +6,29 @@
 - name: Wait for elasticsearch to startup
   wait_for: host={{es_api_host}} port={{es_api_port}} delay=10
 
-- name: Get template files
-  find: paths="/etc/elasticsearch/templates" patterns="*.json"
-  register: templates
-
 - name: Install templates without auth
   uri:
-    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item.path | filename}}"
+    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item | filename}}"
     method: PUT
     status_code: 200
     body_format: json
-    body: "{{ lookup('file', item.path) }}"
+    body: "{{ lookup('file', item) }}"
   when: not es_enable_xpack or not es_xpack_features is defined or "security" not in es_xpack_features
-  with_items: "{{ templates.files }}"
+  with_fileglob:
+    - "{{ es_templates_fileglob | default('') }}"
+  run_once: True
 
 - name: Install templates with auth
   uri:
-    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item.path | filename}}"
+    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item | filename}}"
     method: PUT
     status_code: 200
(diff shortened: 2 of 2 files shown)


barryib:
> Fixes an issue loading installed templates into ES. The use of file lookup meant that the template files are checked on the machine that initiated the Ansible play, this only works if Ansible has been run locally, breaking for remote machines.

@lobsterdore I think the intent of those tasks is to distribute (copy and load) templates which are located on the Ansible server.

Furthermore, I think that using `slurp` module will copy the template content from remotes host to the Ansible server before sending back to the remote server in the body of the `uri`.

Maybe using only `with_fileglob` in the handler like [here](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-templates.yml#L10) and [there](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-templates.yml#L5) (when `es_templates_fileglob` is not defined) is a good trade-off. In that way, we only load template which are on the Ansible server and not those are loaded by mistake on the Elasticsearch nodes.

welchwilmerck:
My understanding of Elasticsearch is that template files only mean something when installed into the cluster via the REST API. In other words, it doesn't matter to ES how many copies or versions of template files exist or what file systems they're on. All that matters is the content of the last one inserted into ES under a given name. As every cluster member supports the API, submitting the same template file to each member as Ansible rotates through them is redundant, but is idempotent.

Someone please correct, with specifics, if this isn't the case.

My understanding of Ansible is that it always runs on only one machine - a control (local) machine -with the desired state of other (remote) machines encoded in, generally, YAML. No machine is referred to as an Ansible server in the Ansible documentation. The control machine pushes Ansible generated code to the remote machines to bring them to that state. The remote machine don't run Ansible, only the generated code. Reaching out to remote machines to pull back state (a configuration file, for example) for propagation elsewhere seems backwards. In particular, what is the provenance of said state? If there are different versions on different remote machines, how do you choose?

Suppose my understanding of template files is incomplete and they need to be on the file system of each cluster member as well as installed into the cluster. That further supports that they should be pushed, not slurped.

I believe this PR doesn't address the issue and may corrupt existing clusters.

barryib:
@welchwilmerck

> My understanding of Elasticsearch is that template files only mean something when installed into the cluster via the REST API. In other words, it doesn't matter to ES how many copies or versions of template files exist or what file systems they're on. All that matters is the content of the last one inserted into ES under a given name. As every cluster member supports the API, submitting the same (local) template file to each member's API endpoint as Ansible rotates through them is redundant, but is idempotent.

Right. Submitting the elasticsearch template to the API endpoint can be be done once with `run_once`, but copying those templates on each node must be done to get some consistency during the deployment.

> My understanding of Ansible is that it always runs on only one machine - a control (local) machine -with the desired state of other (remote) machines encoded in, generally, YAML. No machine is referred to as an Ansible server in the Ansible documentation.

Your definition is right. As you mention ansible doesn't need to be installed on remote hosts (here elasticsearch nodes).

What I ment is to get deployment consistency, you have to put the elasticsearch templates on your Ansible server. Per default into the role `files/templates/`. 

When you run your playbook, ansible will

1. Copy templates on all your nodes (only for deployment consistency)
```yml
- name: Copy templates to elasticsearch
  copy: src={{ item }} dest=/etc/elasticsearch/templates owner={{ es_user }} group={{ es_group }}
  notify: load-templates
  with_fileglob:
    - "{{ es_templates_fileglob | default('') }}"
```

2. Load templates with an API call (with the Ansible handler)
```yml
- name: Install templates without auth
  uri:
    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item.path | filename}}"
    method: PUT
    status_code: 200
    body_format: json
    body: "{{ lookup('file', item.path) }}"
  when: not es_enable_xpack or not es_xpack_features is defined or "security" not in es_xpack_features
  with_fileglob:
    - "{{ es_templates_fileglob | default('') }}"
  run_once: True
```
__NOTE:__ As you notice, I added `run_once` to call the API on only one node.

No need to use `slurp` to copy file content (you already have) on the Ansible server before doing an API call.

welchwilmerck:
A couple of alternatives to copying every template file to every remote simply to detect changes to templates between runs of the playbook.
- Forget the 'build' avoidance and always run_once install all ES templates.
- Cache the 'last installed' on just the control machine, not every remote. Make the copy run_once, but to some (version controlled) directory on the control machine.

lobsterdore:
@barryib I've implemented your fileglob suggestion, I've tested this and it's all working fine for me.

@welchwilmerck I think any sort of refactor should be captured as an issue and handled elsewhere, this is just a simple bug fix PR :)

handlers/main.yml
```diff
@@ -13,3 +7,11 @@
 - name: load-templates
   include: ./handlers/elasticsearch-templates.yml
   when: es_templates
+
```
gingerwizard: This needs to occur before the load templates i think - for the reasons state in the above comment.

gingerwizard: i see https://github.com/ansible/ansible/issues/20603

gingerwizard:
Apologies for the delay here, I had stepped away from the role for a few months due to commitments but elastic are committed to ensuring this is resolved.
I'm reviewing this now. This is causing the template tests to fail so its defn an issue thanks @lobsterdore

So as far as i can see, we have a number of issues here:

1. We are being impacted by this https://github.com/ansible/ansible/issues/20603, which right now is causing the handlers to fail to even load.
1. We copy the templates to the server - purely to detect if they change and thus need to fire the handler. 
1. We, however, then use a lookup - this runs on the local machine, causing the load to actually fail in reality. Our tests passed as the playbook actually runs local to the local docker instance - they are thus present, having been copied above.

So i believe this PR resolves this the lookup problem - however, it changes the order of the handler fire and restart. This is going to cause issues if security is removed for example.
I tempted to move the restart to [here](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/main.yml#L44) and otherwise merge these changes.

gingerwizard:
I'm going merge as part of a larger PR fixing issues with ES not starting. Expect shortly. Thanks to @lobsterdore and @barryib for their patience.

barryib:
@gingerwizard +1

gingerwizard:
Merged through https://github.com/elastic/ansible-elasticsearch/pull/358

PR #301, opened 2017-05-09
Cleanup installation and make shared file system prefixes optional

bndabbs:
This is a lot of small changes and I wanted to submit them as separate pull requests for easier review, but a lot of them ended up being dependent on each other. I believe ee73f18 can be pulled out and moved to a separate PR if necessary, but I left it in there because I couldn't get my playbook to run without it.

The primary goal of this PR is to make the shared filesystem stuff optional, because I got tired of having to prefix the instance name to the service name when performing maintenance. I think the prefix is useful and necessary when operating in specific environments, but it adds unnecessary complexity for a lot of use cases.

### DIFF
--- a/tasks/elasticsearch-scripts.yml
+++ b/tasks/elasticsearch-scripts.yml
@@ -1,6 +1,6 @@
 ---
 
-- set_fact: es_script_dir={{ es_conf_dir }}/{{es_instance_name}}
+- set_fact: es_script_dir={{ es_conf_dir }}{{ es_conf_dir }}{{'/' + es_instance_name if es_use_shared_fs else ''}}
   tags:
       - always
 
--- a/tasks/elasticsearch-scripts.yml
+++ b/tasks/elasticsearch-scripts.yml
@@ -1,6 +1,6 @@
 ---
 
-- set_fact: es_script_dir={{ es_conf_dir }}{{'/' + es_instance_name if es_use_shared_fs else ''}}
+- set_fact: es_script_dir={{ es_conf_dir }}/scripts{{'/' + es_instance_name if es_use_shared_fs else ''}}
   tags:
       - always
 
--- a/tasks/elasticsearch-scripts.yml
+++ b/tasks/elasticsearch-scripts.yml
@@ -1,6 +1,6 @@
 ---
 
-- set_fact: es_script_dir={{ es_conf_dir }}{{ es_conf_dir }}{{'/' + es_instance_name if es_use_shared_fs else ''}}
+- set_fact: es_script_dir={{ es_conf_dir }}{{'/' + es_instance_name if es_use_shared_fs else ''}}
   tags:
       - always
 
--- a/templates/elasticsearch.yml.j2
+++ b/templates/elasticsearch.yml.j2
@@ -8,7 +8,7 @@ cluster.name: elasticsearch
 {% endif %}
 
 {% if es_config['node.name'] is not defined %}
-node.name: {{inventory_hostname}}{{'-' + es_instance_name if es_use_shared_fs else ''}}
+node.name: {{ es_node_name }}
 {% endif %}
 
 #################################### Paths ####################################
--- a/templates/elasticsearch.yml.j2
+++ b/templates/elasticsearch.yml.j2
@@ -8,7 +8,7 @@ cluster.name: elasticsearch
 {% endif %}
 
 {% if es_config['node.name'] is not defined %}
-node.name: {{inventory_hostname}}-{{es_instance_name}}
+node.name: {{inventory_hostname}}{{' - ' + es_instance_name if es_use_shared_fs else ''}}
 {% endif %}
 
 #################################### Paths ####################################
--- a/templates/init/redhat/elasticsearch.j2
+++ b/templates/init/redhat/elasticsearch.j2
@@ -67,7 +67,7 @@ if [ ! -z "$CONF_FILE" ]; then
 fi
 
 exec="$ES_HOME/bin/elasticsearch"
-prog="{{es_instance_name + '_' if es_use_shared_fs else ''}}{{default_file | basename}}"
+prog="{{instance_default_file | basename}}"
 pidfile="$PID_DIR/${prog}.pid"
 
 export ES_JAVA_OPTS
--- a/templates/init/redhat/elasticsearch.j2
+++ b/templates/init/redhat/elasticsearch.j2
@@ -67,7 +67,7 @@ if [ ! -z "$CONF_FILE" ]; then
 fi
 
 exec="$ES_HOME/bin/elasticsearch"
-prog="{{es_instance_name}}_{{default_file | basename}}"
+prog="{{es_instance_name + '_' if es_use_shared_fs else ''}}{{default_file | basename}}"
 pidfile="$PID_DIR/${prog}.pid"
 
 export ES_JAVA_OPTS
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -24,6 +24,7 @@ es_use_shared_fs: false
 es_pid_dir: "/var/run/elasticsearch"
 es_data_dirs: ['/var/lib/elasticsearch']
 es_log_dir: "/var/log/elasticsearch"
+es_conf_dir: "/etc/elasticsearch"
 es_max_open_files: 65536
 es_max_threads: "{{ 2048 if ( es_version | version_compare('6.0.0', '<')) else 8192 }}"
 es_max_map_count: 262144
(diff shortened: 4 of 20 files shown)


Slach:
es_use_shared_fs: false
it's a really usefull feature
pls someone, merge this PR

greenhua:
good! +1

gingerwizard:
Other than the plugin handling (which i think has subsequently improved) this makes considerable improvements to the current role. I'd like to get this merged but i think we first need to fix some critical issues this doesn't resolve. @barryib any thoughts on this? The ability to have only one instance per machine, and thus not be impacted by the `instance_name` seems like a popular request.

kireevco:
Any chance to take care of this? Thanks!

bndabbs:
This PR had gotten pretty far behind the current master, so I did a rebase this evening to bring everything up-to-date.

kireevco:
@elasticmachine test (?)
@gingerwizard any additional changes before this can be merged?

Thank you

kireevco:
@bndabbs actually, I just realized that `es_instance_name` is still appended to everything. Is there a reason why we still keep it in the single-node mode? if it is not specified ES breaks:
`Starting Elasticsearch-...`

kireevco:
I submitted changes back to your fork @bndabbs https://github.com/bndabbs/ansible-elasticsearch/pull/1

ajrpayne:
I submitted changes that I believe clean up the params section a bit. @bndabbs

karmi:
Hi @bndabbs, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/301.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

bndabbs:
@karmi I believe the emails are sorted out now.

tasks/elasticsearch-scripts.yml
```diff
@@ -1,6 +1,6 @@
 ---
 
-- set_fact: es_script_dir={{ es_conf_dir }}/{{es_instance_name}}
+- set_fact: es_script_dir={{ es_conf_dir }}{{ es_conf_dir }}{{'/' + es_instance_name if es_use_shared_fs else ''}}
```
janhoy: The double `es_conf_dir` looks like a bug?

ajrpayne: It does to me. I made a pull request with @bndabbs to remove it.

karmi:
@bndabbs, yes, thanks, the check is now green!

ajrpayne:
I think es_use_shared_fs: true should be the default to maintain backwards-compatibility.

tasks/elasticsearch-scripts.yml
bndabbs: This has been merged now.

bndabbs:
@ajrpayne good call. I flipped the default on that.

janhoy:
I’ve tested your PR with my install and it looks good. Is there perhaps need for README changes as well, before merging?

janhoy:
@bndabbs, @karmi What's left on this now?

janhoy:
@bndabbs We think we might have found a bug with this PR. If you have plugins defined, upgrading to a newer ES version using this role gives this error:

    /usr/share/elasticsearch/bin/elasticsearch -Epath.conf=/etc/elasticsearch
    [2017-11-27T14:52:22,712][WARN ][o.e.b.ElasticsearchUncaughtExceptionHandler] [control] uncaught exception in thread [main]
    org.elasticsearch.bootstrap.StartupException: java.lang.IllegalArgumentException: plugin [analysis-phonetic] is incompatible with version [5.6.4]; was designed for version [5.6.3]
	at org.elasticsearch.bootstrap.Elasticsearch.init(Elasticsearch.java:136)
        ...

I believe the problem is that the variable `es_install.changed` is used to determine whether to reinstall plugins but the var is never defined anywhere...?

janhoy:
Well, I now understand that `es_install.changed` gets automatically set if the install task causes a change in state. But it does not seem to work, the var `es_plugins_reinstall` gets set to False in line 6 of `elasticsearch-plugins.yml`:

    set_fact: es_plugins_reinstall={{ es_install.changed }}

I can workaround the bug by setting that line to `True`. It does *not* help to set the `es_plugins_reinstall` variable to True in my inventory.

janhoy:
Hello @bndabbs ? :)

bndabbs:
@janhoy I'll take a look at the plugin logic and see if I can figure that out.

janhoy:
@bndabbs figured it out yet? I'd really like for this PR to be merged soon as we base our install on it.

janhoy:
Hi, any luck with this? I'm considering reverting back to latest release since this PR seems to be going nowhere anytime soon?

ajrpayne:
@janhoy 
What OS are you using? Was 5.6.4 installed by some other method that would cause es_install to return false?

janhoy:
I've rolled back to default version of the role since this PR is going nowhere. I was using CentOS, and installed with this role+PR from a clean system. It is easily reproducible by doing an install with plugins, then do an upgrade to a newer minor version.

Crazybus:
Hi @bndabbs!

This looks really great! I ran out of time today but tomorrow I'm going to have a good look over it and leave a proper review. 

Could you rebase this against master please? Then we can run this against the automated testing job to see if it is breaking anything for `5.x` and `6.x`

Cheers,

Michael

ajrpayne:
I think I see the issue @janhoy had @bndabbs.

If installing when es_use_repository is true the task to install when es_use_repository is false, even when skipped, sets es_install.change to false.

ajrpayne:
default: ok: [dev1-bpv01] => {
    default:     "msg": {
    default:         "changed": false, 
    default:         "skip_reason": "Conditional result was False", 
    default:         "skipped": true
    default:     }
    default: }

ajrpayne:
I may a pull request for @bndabbs that should fix the plugin logic issue.

templates/elasticsearch.yml.j2
```diff
@@ -8,7 +8,7 @@ cluster.name: elasticsearch
 {% endif %}
 
 {% if es_config['node.name'] is not defined %}
-node.name: {{inventory_hostname}}-{{es_instance_name}}
+node.name: {{inventory_hostname}}{{' - ' + es_instance_name if es_use_shared_fs else ''}}
```
Crazybus: Are the extra spaces around ` - ` intentional? This is causing the tests to fail because the node changed `localhost-node1` to `localhost - node1`

templates/init/redhat/elasticsearch.j2
```diff
@@ -67,7 +67,7 @@ if [ ! -z "$CONF_FILE" ]; then
 fi
 
 exec="$ES_HOME/bin/elasticsearch"
-prog="{{es_instance_name}}_{{default_file | basename}}"
+prog="{{es_instance_name + '_' if es_use_shared_fs else ''}}{{default_file | basename}}"
```
Crazybus: Would it be possible to move all of `'_' if es_use_shared_fs else ''` logic into `defaults/main.yml`? It makes it a lot more readable and means that any future changes won't need to keep copy pasting this if else statement everywhere.

ajrpayne:
I like the changes @bndabbs. Made a pull request that I thought brought it more in line with the sharedfs structure you were going for.

kireevco:
@bndabbs Any chance to make a last push to get it merged? Thanks!

rozling:
Would love to see this merged.  Is there anything I / others can do to help?

Slach:
so sad ;-(( this PR was really usefull i hope currently ansible-elasticsearch role implements some analogs

jmlrt:
Hi @Slach,
What exact part of this PR did you need?
Don't hesitate to create a new PR based on current master branch if you think there is still a need for it.

Slach:
i use this PR for install once elasticsearch instance to once host directly to /var/lib/elasticsearch/data instead of /var/lib/something-with-instance-name/

jmlrt:
Are you still using this 2 years old branch PR for your Elasticsearch deployments?
I think what you need has already been implemented in https://github.com/elastic/ansible-elasticsearch/pull/566. You can refer to [multi-instance.md](https://github.com/elastic/ansible-elasticsearch/blob/master/docs/multi-instance.md) and https://github.com/elastic/ansible-elasticsearch/issues/554#issuecomment-496804929 for more details.

I also encourage you to use 7.4.0 release [Ansible Galaxy](https://galaxy.ansible.com/elastic/elasticsearch)  or `master` branch of the role now that we are trying to release our roles more often (please check the [CHANGELOG.md] (https://github.com/elastic/ansible-elasticsearch/blob/master/CHANGELOG.md) and test it on a dev environment before as they will be breaking changes).

Slach:
@jmlrt oh, thanks a lot for explanation ;-) 
no, currently i use clickhouse with ngram indexes and Grafana Loke instead of Elasticsearch 
so i happy for this ticket close

Issue #316, opened 2017-06-16
Debian: openjdk-8-jre-headless : Depends: ca-certificates-java but it is not going to be installed


mgodlewski:
Hello,

On debian, I can't install openjdk-8-jre-headless since it has unmet dependencies on ca-certificates

```
TASK [ansible-elasticsearch : Debian - Ensure Java is installed] ***************
fatal: [xxxx]: FAILED! => {"cache_update_time": 1497603840, "cache_updated": false, "changed": false, "failed": true, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"    --simulate install 'openjdk-8-jre-headless'' failed: E: Unable to correct problems, you have held broken packages.\n", "stderr": "E: Unable to correct problems, you have held broken packages.\n", "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nSome packages could not be installed. This may mean that you have\nrequested an impossible situation or if you are using the unstable\ndistribution that some required packages have not yet been created\nor been moved out of Incoming.\nThe following information may help to resolve the situation:\n\nThe following packages have unmet dependencies:\n openjdk-8-jre-headless : Depends: ca-certificates-java but it is not going to be installed\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Some packages could not be installed. This may mean that you have", "requested an impossible situation or if you are using the unstable", "distribution that some required packages have not yet been created", "or been moved out of Incoming.", "The following information may help to resolve the situation:", "", "The following packages have unmet dependencies:", " openjdk-8-jre-headless : Depends: ca-certificates-java but it is not going to be installed"]}
```

Regards,

Mathieu

gingerwizard:
I think we are going to remove management of java. This isn't the purpose of the role and its not supportable longer term. The java component will simply check that its installed and the version is valid. https://github.com/elastic/ansible-elasticsearch/issues/268

PR #317, opened 2017-06-16
install openjdk-8 and ca-certificates from backports

mgodlewski:
This should fix issue #316

### DIFF
--- a/tasks/java.yml
+++ b/tasks/java.yml
@@ -15,7 +15,7 @@
   when: ansible_os_family == 'Debian'
 
 - name: Debian - Ensure Java is installed
-  apt: name={{ java }} state={{java_state}}
+  apt: name={{ java }} state={{java_state}} default_release=jessie-backports
   when: ansible_os_family == 'Debian'
 
 - shell: java -version 2>&1 | grep OpenJDK


gingerwizard:
I don't feel qualified to comment on this. @dliappis maybe?

gingerwizard:
@jpcarey maybe you have some thoughts here?

jpcarey:
@mgodlewski do you have any further supporting information. Both debian8 and 9 seem to have the necessary components to install `openjdk-8-jre-headless` without need for using backports.

@gingerwizard I'd favor removing java installation from this role. If a user wants to use backports (, that should be an external dependency of installing / provisioning elasticsearch.

https://backports.debian.org/Instructions/
> Backports are recompiled packages from testing (mostly) and unstable (in a few cases only, e.g. security updates) in a stable environment so that they will run without new libraries (whenever it is possible) on a Debian stable distribution
> 
> Backports cannot be tested as extensively as Debian stable, and backports are provided on an as-is basis, with risk of incompatibilities with other components in Debian stable. Use with care!

gingerwizard:
@jpcarey the plans is indeed to remove. I propose we close the PR and begin work on removing Java as a priority.

PR #318, opened 2017-06-16
Change basestring by str to make it python3 compatible

viscat:
- Change basestring by str to make it python3 compatible

### DIFF
--- a/filter_plugins/custom.py
+++ b/filter_plugins/custom.py
@@ -13,7 +13,7 @@ def modify_list(values=[], pattern='', replacement='', ignorecase=False):
     return [_re.sub(replacement, value) for value in values]
 
 def append_to_list(values=[], suffix=''):
-    if isinstance(values, basestring):
+    if isinstance(values, str):
         values = values.split(',')
     return [str(value+suffix) for value in values]
 


gingerwizard:
@viscat can you please sign the CLA?

viscat:
Done

gingerwizard:
@viscat this causes the tests to fail. Initial investigation appears to indicate this is the result of they way ansible is passing the value. the type check of str on python 2.7, which is what the tests use seems to fail when "/var/lib/elasticsearch" is passed - this results in the values being treated as a set of chars by the loop `return [str(value+suffix) for value in values]`

Hence path.data becomes

`path.data: //localhost-node1,v/localhost-node1,a/localhost-node1,r/localhost-node1,//localhost-node1,l/localhost-node1,i/localhost-node1,b/localhost-node1,//localhost-node1,e/localhost-node1,l/localhost-node1,a/localhost-node1,s/localhost-node1,t/localhost-node1,i/localhost-node1,c/localhost-node1,s/localhost-node1,e/localhost-node1,a/localhost-node1,r/localhost-node1,c/localhost-node1,h/localhost-node1`

gingerwizard:
@viscat im testing but i think

```
from six import string_types
if isinstance(values, string_types):
```

is robust and will work across python 2 and 3.

gingerwizard:
https://github.com/elastic/ansible-elasticsearch/pull/364
Introduces the above.

gingerwizard:
Thanks @viscat i believe we have a solution through #364. Thankyou for addressing this.

viscat:
Oh, sorry and thank you very much @gingerwizard

PR #331, opened 2017-07-07
Enable certificate deployment as part of X-Pack Security set-up

cdahlqvist:
The playbook does not currently seem to support deploying and configuring certificates as part of setting up X-Pack security. This PR adds support for uploading and deploying certificates as part of the installation. I am new to Ansible, so any feedback and suggestions around how to improve this will be much appreciated.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -25,6 +25,10 @@ es_max_map_count: 262144
 es_allow_downgrades: false
 es_enable_xpack: false
 es_xpack_features: ["alerting","monitoring","graph","security"]
+es_xpack_ssl_key_src: ''
+es_xpack_ssl_certificate_src: ''
+es_xpack_ssl_certificate_authorities_src: ''
+es_xpack_certificates_on_host: 'no'
 #These are used for internal operations performed by ansible.
 #They do not effect the current configuration
 es_api_host: "localhost"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -30,11 +30,6 @@ es_xpack_ssl_certificate_src: ''
 es_xpack_ssl_certificate_authorities_src: ''
 es_xpack_certificates_on_host: 'no'
 
-es_xpack_monitoring_datastore_host: ""
-es_xpack_monitoring_metrics_datastore_port: ""
-es_xpack_monitoring_metrics_datastore_protocol: ""
-es_xpack_monitoring_metrics_datastore_user: ""
-es_xpack_monitoring_metrics_datastore_password: ""
 #These are used for internal operations performed by ansible.
 #They do not effect the current configuration
 es_api_host: "localhost"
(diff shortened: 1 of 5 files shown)


gingerwizard:
@cdahlqvist this is a great addition and something i think @jpcarey planned to add. Can we, however, get a test? I think id want to test that this works on a 2 node deployment min and that the cluster forms.

I assume this test is unlikely to work if we dont provide  ssl.verification_mode:certificate.

 I can help with a test if needed.

jpcarey:
My thought was to just check if http or transport TLS was enabled in the configured elasticsearch.yml settings, then check that the configured cert paths (pem or java keystore type) were valid - if not provide a failure message that certificates had not been deployed to the node being provisioned.  This does get a little tricky in that the certs need to be in the config path, or java security would not allow usage of the files. This has the expectation that the end user's playbook would create the appropriate config folder and place the certs into that location before running this role. I personally think the steps to copy the certs is a config deployment problem outside the scope of this role.

I could however see adding the ability to use the certgen tool we provide for creating and signing certs automatically for each node - but I'd rather see that as a separate role.

gingerwizard:
For reference this functionality is supported in the [puppet module](https://github.com/elastic/puppet-elasticsearch#certificates), although it uses a java keystore. Ansible supports managing keystores in addition.

Whilst i agree with @jpcarey as to the role division, the need for the certs to be in the config path troubles me - especially given we use `instance_name` to support multiple instances on a box - thus the config paths are variable.  Could we maybe ask the user to provide locations on the server, and simply copy them into the required locations ourselves. I think this would be reasonable compromise.

@jpcarey

jpcarey:
@gingerwizard so end user would provide an array of files (generic, to support any type), and we would just make sure each is copied?

gingerwizard:
@jpcarey the array they provide would be files on the remote machine - rather than the local machine.

gingerwizard:
And yes we we ensure they are copied into each instance and the parameters inserted into the elasticsearch.yml

gingerwizard:
@jpcarey do we have an agreed approach here? I think its pretty simple and id like to try and get it done before 6.0

medcl:
I'd like to get feature this merge into master, is there any update?

rozling:
I would also love to see this merged soon.

However unless I'm mistaken there is one step missing to enable certificate deployment (from the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html#tls-transport)):

> If you secured the node’s certificate with a password, add the password to your Elasticsearch keystore:
> [snip]

If this isn't done, Elasticsearch doesn't start as it can't read the certificate, and the playbook hangs.

> Whilst i agree with @jpcarey as to the role division, the need for the certs to be in the config path troubles me - especially given we use `instance_name` to support multiple instances on a box - thus the config paths are variable

@gingerwizard now that it looks like https://github.com/elastic/ansible-elasticsearch/pull/566 will be merged soon, maybe this simplifies things a bit?

jmlrt:
I'm closing this PR as we have current work in progress in https://github.com/elastic/ansible-elasticsearch/pull/597

PR #347, opened 2017-08-04
override log4j2.properties with custom template

phamhongviet:
be able to override log4j2.properties with custom template using variable es_config_log4j2 as the template path


gingerwizard:
Merging. Ill make a separate PR to ensure we have a basic test for this.

gingerwizard:
Tested under https://github.com/elastic/ansible-elasticsearch/pull/361

PR #350, opened 2017-08-10
Remove default.path.home argument from version 5.4 and above

carlosfunk:
Removes default.path.home as it is no longer supported from version 5.4

### DIFF
--- a/templates/init/debian/elasticsearch.j2
+++ b/templates/init/debian/elasticsearch.j2
@@ -87,7 +87,12 @@ fi
 # Define other required variables
 PID_FILE="$PID_DIR/$NAME.pid"
 DAEMON={{es_home}}/bin/elasticsearch
+
+{% if es_version | version_compare('5.4', '<') %}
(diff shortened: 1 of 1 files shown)


karmi:
Hi @carlosfunk, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/350.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

carlosfunk:
Ah that was an old company email, I've pushed again with the correct one.

gingerwizard:
Actually just saw this apologies @carlosfunk. In updating to 5.5.1 under https://github.com/elastic/ansible-elasticsearch/pull/358 we will fix this. Ill add this PR to that one for credit.

gingerwizard:
Closed by https://github.com/elastic/ansible-elasticsearch/pull/358

PR #358, opened 2017-08-16
Fixes templates, ES hang and upgrades to 5.5.1

gingerwizard:
This makes a number of changes:

1. Upgrades us to 5.5.1. Syncs new initd, systemd and jvm options file. The initd differences were preventing ES from starting on 5.3+. This is now resolved.
2. Templates have been reworked. @lobsterdore provided the fix here - i merged in #300. Notice we now don't use handlers due to reliability concerns - we still copy the templates to the box but invoke using with_file_glob. I would like to make the request using the files on the box ideally but this works for now. These tests pass. @barryib please take a look.
3. I've had a lot of issues with Ansible versions - here we enforce v 2.3.2.0.  Pip is now used in the provisioning to ensure this is the version on the test docker containers . We should not be able to test against specific versions.

What this doesn't fix:

1. Issues with x-pack. This is broken when adding file realm users. I'm looking at this next. Expect a PR. RESOLVED - docker issue.
2. Intermittent issues with the kitchen framework and net-scp failing to copy files to the docker containers. I'm investigating but think i've resolved - gem conflict issues. 

I don't expect the automated tests to pass yet so i'm going to merge this in and work on them after the x-pack issue.


tasks/main.yml
```diff
@@ -37,24 +37,30 @@
...
 
-- include: elasticsearch-templates.yml
+- include: elasticsearch-template-copy.yml
   when: es_templates
   tags:
       - templates
 
 - meta: flush_handlers
 
+- include: elasticsearch-template-insert.yml
```
gingerwizard: Notice we flush the handlers before doing the insert. This is to ensure a restart occurs if needed. Suppose the user makes a config change which impacts the ability to insert templates - the first one that springs to mind is turning off security. Admittedly this is a niche case.

test/integration/xpack-limited.yml
```diff
@@ -0,0 +1 @@
+#no security
```
gingerwizard: Yes this file was placeholder for a new set of tests - x-pack without security. ill remove and repush.

handlers/main.yml
```diff
@@ -1,19 +1,13 @@
...
+
 - name: restart elasticsearch
   service: name={{instance_init_script | basename}} state=restarted enabled=yes
-  when: 
-    - es_restart_on_change 
-    - es_start_service 
+  when:
+    - es_restart_on_change
+    - es_start_service
     - ((plugin_installed is defined and plugin_installed.changed) or (config_updated is defined and config_updated.changed) or (xpack_state.changed) or (debian_elasticsearch_install_from_repo.changed or redhat_elasticsearch_install_from_repo.changed or elasticsearch_install_from_package.changed))
```
gingerwizard: They have legacy but avoid unnecessary restarts. I'm hesitant to change them as part of this PR but they are defn need of a review - once we're confident the test suite is passing and sufficiently comprehensive.

defaults/main.yml
```diff
@@ -1,8 +1,9 @@
 ---
 es_major_version: "5.x"
-es_version: "5.2.2"
+es_version: "5.5.1"
 es_version_lock: false
 es_use_repository: true
+es_templates_fileglob: "./files/templates/*"
```
gingerwizard: +1
changing to `files/templates/*.json`

tasks/elasticsearch-template-copy.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+- file: path=/etc/elasticsearch/templates state=directory owner={{ es_user }} group={{ es_group }}
+
+- name: Copy templates to elasticsearch
+  copy: src={{ item }} dest=/etc/elasticsearch/templates owner={{ es_user }} group={{ es_group }}
+  register: load_templates
```
gingerwizard: So this variable is set by the template-copy. It confirms we need to do a insert as they have changed.

gingerwizard:
@barryib i believe i've addressed most of your concerns.

tasks/main.yml
barryib: I thought that loading templates was an idempotent task for elasticsearch, so no need to handle this in 2 steps.

By the way, if you want to load only if the copy happen, I think you have to use `load_templates|changed`

tasks/elasticsearch-template-insert.yml
```diff
@@ -0,0 +1,36 @@
...
+  service: name={{instance_init_script | basename}} state=started enabled=yes
+  when: es_start_service
+
+- name: Wait for elasticsearch to startup
+  wait_for: host={{es_api_host}} port={{es_api_port}} delay=10
+  when: es_start_service
+
+- name: Install templates without auth
+  uri:
+    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item | filename}}"
```
barryib: @gingerwizard if we use filename with `.json` extension, should we leave it in the template name (in the url)?

Tips: Since Ansible 2.0 `{{ (item | filename | splitext)[0] }}`

tasks/main.yml
```diff
@@ -37,24 +37,30 @@
   tags:
       - xpack
 
-- include: elasticsearch-templates.yml
+- include: elasticsearch-template-copy.yml
```
gingerwizard: See below.

tasks/elasticsearch-template-insert.yml
gingerwizard: The inserted template should be the basename of the file (minus the extension). This should be sufficiently unique as it what users would expect no?

gingerwizard:
This https://github.com/elastic/ansible-elasticsearch/pull/350 is also fixed by this PR as it updates the init files to the latest version. Thanks to @carlosfunk who picked this up.

tasks/elasticsearch-template-insert.yml
barryib: yep, agree. Specially as users would expect :)

handlers/main.yml
barryib: agree.

tasks/elasticsearch-template-insert.yml
gingerwizard: I'm not clear how these tests are passing then as we test for the filename. Let me check this and ill update the PR if its broken or the test is broken.

gingerwizard: So we insert with the basename already. The | filename filter is actually overriden [here](https://github.com/elastic/ansible-elasticsearch/blob/master/filter_plugins/custom.py#L31)

Happy to accept PRs refactoring this but this works as intended.

tasks/main.yml
gingerwizard: ah yes good spot. We don't have an idempotent test for this - im going to add one now in the standard tests and update the PR.  I originally intended the template insert file to be where we would do any actions after restarting, but i think we can do the restart and then copy the files and do the insert. I'll update and test.

tasks/elasticsearch-template-insert.yml
barryib: I didn't see that one. 👍

gingerwizard:
@barryib i have merged the two template files as discussed. We also do an idempotent test on the templates using the `load_templates.changed` variable. The multi, standard,config and package tests  now all pass. 
Please review and LGTM if your satisfied with the changes.

tasks/elasticsearch-template.yml
```diff
@@ -0,0 +1,45 @@
+---
```
gingerwizard: @barryib single file now.

gingerwizard:
@barryib i believe i have addressed all of your concerns and reduced the templates file to a single file. I'm merging this so we can keep the role moving forward - we still aren't green on all tests and need to set up a new automated test environment.

tasks/elasticsearch-template.yml
```diff
@@ -0,0 +1,45 @@
...
+  when: es_start_service and load_templates.changed
+
+- name: Install templates without auth
+  uri:
+    url: "http://{{es_api_host}}:{{es_api_port}}/_template/{{item | filename}}"
+    method: PUT
+    status_code: 200
+    body_format: json
+    body: "{{ lookup('file', item) }}"
+  when: load_templates.changed and es_start_service and not es_enable_xpack or not es_xpack_features is defined or "security" not in es_xpack_features
```
barryib: @gingerwizard Is using directly `load_templates.changed` ok? I'm afraid that `changed` key won't exist if nothing change => then this task will fail.

I think using the filter (`load_templates | changed`) prevent you that.

Can you check please? Maybe, i'm wrong

barryib:
@gingerwizard I've only one last question about `load_templates.changed`.

If ok, everything else LGTM.

Thanks again for all your great work.

PR #364, opened 2017-08-18
Python3 compatibilty

gingerwizard:
Makes the role both python 2 and python 3 compatible. This relies on the `six`module to reliably evaluate whether strings or arrays are passed for paths. Inspired by @viscat through https://github.com/elastic/ansible-elasticsearch/commit/0eadf596eb60230b0d1e18f4e435a7b878037233


jpcarey:
LGTM

Issue #315, opened 2017-06-10
Elasticsearch startup times out on Centos 7 in Vagrant


tonyrush:
Using the simplest playbook that doesn't error out because the user must be root and the ansible_local provisioner:
---
- name: Simple Example
  hosts: localhost
  remote_user: root
  become: yes
  become_method: sudo
  roles:
    - { role: elasticsearch, es_instance_name: "node1" } 
vars:

Fails with:

 [Wait for elasticsearch to startup] ***********************
fatal: [localhost]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for localhost:9200"}
	to retry, use: --limit @/vagrant/simple.retry

PLAY RECAP *********************************************************************
localhost                  : ok=39   changed=19   unreachable=0    failed=1   

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
s_instance_name: "node1" }

rhinoman:
I'm seeing the same thing on CentOS 7 on Linode.  But elasticsearch *is* running according to systemctl status

gingerwizard:
Reproducible on latest version of ES. Investigating.

gingerwizard:
Addressed in https://github.com/elastic/ansible-elasticsearch/pull/366

Issue #305, opened 2017-05-11
Install hangs on [Wait for elasticsearch to startup] Ubuntu 16


runtman:
It seems the installation of elasticsearch via this ansible playbook hands on [Wait for elasticsearch to startup].

If I was to take a stab in the dark, I believe it relates to systemd missing the startup.

`bin/sh -c '"'"'"'"'"'"'"'"'echo BECOME-SUCCESS-iolgltkwkkyiyzdtbcvhnhnejczxkbup; /usr/bin/python /home/ubuntu/.ansible/tmp/ansible-tmp-1494493145.89-115538618345757/wait_for.py; rm -rf "/home/ubuntu/.ansible/tmp/ansible-tmp-1494493145.89-115538618345757/" > /dev/null 2>&1'"'"'"'"'"'"'"'"' && sleep 0'"'"''
[...]

PR #366, opened 2017-08-20
Simpler restart handling + issue-test for recreation of problems

gingerwizard:
This PR simplifies the restart logic, relying on the restart handler only. We now notify the handler consistently and remove the additional checks.

This should close the following issues:

https://github.com/elastic/ansible-elasticsearch/issues/305
https://github.com/elastic/ansible-elasticsearch/issues/315

All tests pass locally. 

I have also added issue-test.yml.  This test will not be included in the new automated tests, but allows us to easily test playbooks for user issues. I'll add instructions once we have the automated test suite - users can in theory then issue a PR with a playbook in the test-issue.yml and immediately demonstrate the failure.

@barryib @jpcarey i believe this simplifies restarts significantly.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -111,3 +111,9 @@ suites:
     provisioner:
       playbook: test/integration/xpack-standard.yml
       idempotency_test: true
+  - name: issue-test
+    run_list:
+    attributes:
+    provisioner:
+      playbook: test/integration/issue-test.yml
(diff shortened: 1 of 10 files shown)


jpcarey:
LGTM.

Haven't looked, but hopefully the notify handler could be overridden with a proper rolling restart at some point.

PR #374, opened 2017-09-06
Plugin installation fixes and install from url 

orphaner:
Hello,
I had a hard time making the plugin installation working as expected. Here are some small fixes + a new feature to install a plugin from an url.

Thanks,
Nicolas

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -46,6 +46,7 @@
   environment:
     CONF_DIR: "{{ conf_dir }}"
     ES_INCLUDE: "{{ instance_default_file }}"
+    ES_ENV_FILE: "{{ instance_default_file }}"
 
 - name: Install elasticsearch plugins
   command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.plugin }} --batch --silent"
@@ -58,6 +59,7 @@
   environment:
     CONF_DIR: "{{ conf_dir }}"
     ES_INCLUDE: "{{ instance_default_file }}"
+    ES_ENV_FILE: "{{ instance_default_file }}"
     ES_JAVA_OPTS: "{% if item.proxy_host is defined and item.proxy_host != '' and item.proxy_port is defined and item.proxy_port != ''%} -Dhttp.proxyHost={{ item.proxy_host }} -Dhttp.proxyPort={{ item.proxy_port }} -Dhttps.proxyHost={{ item.proxy_host }} -Dhttps.proxyPort={{ item.proxy_port }}  {% elif es_proxy_host is defined and es_proxy_host != '' %} -Dhttp.proxyHost={{ es_proxy_host }} -Dhttp.proxyPort={{ es_proxy_port }} -Dhttps.proxyHost={{ es_proxy_host }} -Dhttps.proxyPort={{ es_proxy_port }} {% endif %}"
   until: plugin_installed.rc == 0
   retries: 5
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -54,7 +54,7 @@
   failed_when: "'ERROR' in plugin_installed.stdout"
   changed_when: plugin_installed.rc == 0
   with_items: "{{ es_plugins }}"
-  when: item.plugin in plugins_to_install
+  when: item.plugin is defined and item.plugin in plugins_to_install
   notify: restart elasticsearch
   environment:
     CONF_DIR: "{{ conf_dir }}"
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -49,7 +49,7 @@
     ES_ENV_FILE: "{{ instance_default_file }}"
 
 - name: Install elasticsearch plugins
-  command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.plugin }} --batch --silent"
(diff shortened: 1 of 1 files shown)


tasks/elasticsearch-plugins.yml
```diff
@@ -46,18 +46,20 @@
...
   failed_when: "'ERROR' in plugin_installed.stdout"
   changed_when: plugin_installed.rc == 0
   with_items: "{{ es_plugins }}"
-  when: item.plugin in plugins_to_install
+  when: item.plugin is defined and item.plugin in plugins_to_install
   notify: restart elasticsearch
   environment:
     CONF_DIR: "{{ conf_dir }}"
     ES_INCLUDE: "{{ instance_default_file }}"
+    ES_ENV_FILE: "{{ instance_default_file }}"
```
barryib: @orphaner why do you need the `ES_ENV_FILE`? `CONF_DIR` is not enough?

tasks/elasticsearch-plugins.yml
```diff
@@ -46,18 +46,20 @@
...
 
 - name: Install elasticsearch plugins
-  command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.plugin }} --batch --silent"
+  command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.url | default(item.plugin) }} --batch --silent"
   register: plugin_installed
   failed_when: "'ERROR' in plugin_installed.stdout"
   changed_when: plugin_installed.rc == 0
   with_items: "{{ es_plugins }}"
-  when: item.plugin in plugins_to_install
+  when: item.plugin is defined and item.plugin in plugins_to_install
```
barryib: @orphaner I think it's better to add default value to avoid looping if empty.

```yml
  with_items: "{{ es_plugins | default([]) }}"
  when: item.plugin in plugins_to_install
```

tasks/elasticsearch-plugins.yml
```diff
@@ -7,7 +7,7 @@
 
 - set_fact: list_command=""
 #If we are reinstalling all plugins, e.g. to a version change, we need to remove all plugins (inc. x-pack) to install any plugins. Otherwise we don't consider x-pack so the role stays idempotent.
-- set_fact: list_command="| grep -vE 'x-pack'"
+- set_fact: list_command="| grep -vE 'x-pack' || :"
```
barryib: smart

barryib:
@gingerwizard ping

gingerwizard:
Reviewing now.

tasks/elasticsearch-plugins.yml
```diff
@@ -46,18 +46,20 @@
   environment:
     CONF_DIR: "{{ conf_dir }}"
     ES_INCLUDE: "{{ instance_default_file }}"
+    ES_ENV_FILE: "{{ instance_default_file }}"
```
gingerwizard: I need to check we need this. Against which version of ES did you test this @orphaner ?

orphaner: 5.4.3

tasks/elasticsearch-plugins.yml
gingerwizard: Trying to identify what the supported way forward here. It seems both are supported i.e.

```
# The default env file is defined at building/packaging time.
# For a distribution package, the value is "/etc/default/elasticsearch".
ES_ENV_FILE="/etc/default/elasticsearch"
# If an include is specified with the ES_INCLUDE environment variable, use it
if [ -n "$ES_INCLUDE" ]; then
    ES_ENV_FILE="$ES_INCLUDE"
fi
```
however, id like to use the one that will be supported moving forward.

tasks/elasticsearch-plugins.yml
gingerwizard: @barryib i think we need the additional check incase people forget to specify the plugin var so...

```
with_items: "{{ es_plugins | default([]) }}"
when: item.plugin is defined and item.plugin in plugins_to_install
```
?

tasks/elasticsearch-plugins.yml
```diff
@@ -46,18 +46,20 @@
...
 
 - name: Install elasticsearch plugins
-  command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.plugin }} --batch --silent"
+  command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.url | default(item.plugin) }} --batch --silent"
   register: plugin_installed
```
gingerwizard: although the use of `url` is clearer, this does currently work if the user sets the plugin to a url explicitly. Given it doesn't break backwards compatibility, and its cleaner, happy to merge. @barryib

gingerwizard:
This looks good. I'd like to check on the env variable with the ES team and enhance the loop to avoid unnecessary iteration. We'll also need a README update for the url change also.

tasks/elasticsearch-plugins.yml
gingerwizard: https://github.com/elastic/elasticsearch/pull/25804/files

gingerwizard:
I think we remove `ES_INCLUDE` in favor of `ES_ENV_FILE` - see https://github.com/elastic/elasticsearch/pull/25804/files

@orphaner do you mind adjusting the PR with the changes above and ill test locally - still waiting  on build infra unfortunately.

tasks/elasticsearch-plugins.yml
barryib: @gingerwizard Right. So, LGTM

PR #377, opened 2017-09-15
set "check_mode: no" when computing installed_plugins so that ansible…

jimrobinson:
set "check_mode: no" when computing installed_plugins so that ansible-playbook can succeed when in check mode (-C)


strootman:
Good call.
This wasn't added until Ansible v2.2.

I'm sure there are other places which could benefit from this.

andyshinn:
It looks like this got added in #487. This PR can be closed.

Issue #369, opened 2017-08-25
Change password of reserved user


saudurehman:
<!

**Describe the feature**:
The playbook should allow to change passwords of reserved/default users.
**Playbook**
```yaml
- hosts: localhost
  roles:
    - {
        role: elasticsearch,
        es_config: 
        { 
          "xpack.security.authc.realms.file1.type": "file",
          "xpack.security.authc.realms.file1.order": 1,
          "xpack.security.authc.realms.native1.type": "native",
          "xpack.security.authc.realms.native1.order": 0
        },
    es_instance_name: "security_node" }
  vars:
    es_heap_size: "1g"
    es_enable_xpack: true
    es_plugins:
      - plugin: ingest-geoip
    es_xpack_features:
      - security
      - alerting
    es_api_basic_auth_username: elastic
    es_api_basic_auth_password: changeme
    es_users:
      file:
        kibana:
          password: changeme
          roles:
            - kibana_system
```

gingerwizard:
Ok i see the issue. We try to add the reserved users which is obviously not possible. Whilst fixable, im concerned about the `elastic` user. If a user changes the user they are making API calls with we will need to detect this and user the correct password subsequently - otherwise calls will fail.

gingerwizard:
There is a couple of things here:

1. We need to document you can't change the reserved users and roles through the file realm. I'm considering explictly ignoring them, but this will require a static list as a param.  I'd prefer we just document. 
2. Support changing through the native realm but password only - this requires us to read the reserved users through a call. Should be simple. Reserved roles should either fail or be ignored. Probably the latter.
3. If the user changes the password of the `es_api_basic_auth_username` we need to do this first and update the param.

gingerwizard:
Closed by #379

DanPhonovation:
Sorry to re-open; I'm slightly unclear after examining the changeset. What is required to change the default of elastic:changeme?

PR #379, opened 2017-09-19
Fix for issue #369

gingerwizard:
This aims to fix issue #369.  A couple of problems were identified:

1. We weren't able to set the password for reserved users e.g. kibana or elastic.  This is now possible through the native realm only. Attempting to do this through the file realm is actually an invalid operation in ES. 

**native realm**
For the native realm we query ES to get the list of users as these are subject to change across versions, and then only allow the password to be set on those that are reserved i.e. role changes are ignored.  No changes are allowed to reserved roles at all.  If the user attempts to change a reserved role, or reserved users membership to a role, in the native realm we simply log a warning rather than hard fail. 

**file realm**
We have the similar behaviour in the file realm. This, however, relies on a listed hard coded reserved_user_names [here](https://github.com/elastic/ansible-elasticsearch/compare/master...gingerwizard:master#diff-6d62b73b3f46cfa11f6466d08cf45303L7) as the file realm is done before ES is started.  I'm unsure of what to do for reserved roles in this realm - we could add a list of reserved roles but this will require maintenance. @barryib @jpcarey any thoughts? I'd happily remove this logic and just document "DON'T CHANGE RESERVED USERS/ROLES IN THE FILE REALM".

2. In providing the above facility we need to handle if the user changes the password of the `es_api_basic_auth_username` - there is some simple logic to address this. In most cases this will be the elastic user so this is a fairly fundamental ability. We set this users password explicitly and change the variable. The user could in theory get themselves into a pickle by using a different user and either not specifying them in the user list (this causing the user to removed) or giving them insufficient permissions to make changes through this user. We don't protect against this and it seems overly complicated to do so. Again @jpcarey @barryib 

@strootman for review also.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -116,4 +116,4 @@ suites:
     attributes:
     provisioner:
       playbook: test/integration/issue-test.yml
-      idempotency_test: true
+      idempotency_test: false
--- a/filter_plugins/custom.py
+++ b/filter_plugins/custom.py
@@ -21,10 +21,10 @@ def append_to_list(values=[], suffix=''):
 def array_to_str(values=[],separator=','):
     return separator.join(values)
 
-def extract_role_users(users={}):
+def extract_role_users(users={},exclude_users=[]):
     role_users=[]
     for user,details in users.iteritems():
-        if "roles" in details:
+        if user not in exclude_users and "roles" in details:
             for role in details["roles"]:
                 role_users.append(role+":"+user)
     return role_users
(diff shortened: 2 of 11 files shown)


gingerwizard:
Oh and i enhanced the tests to check the above.

gingerwizard:
I've pushed some additional changes:

1. Moved the check for use of reserved users in the file realm to the parameters check task. Here it is now a fail as requested by @jpcarey. We hard fail early and prevent even install of ES. I think this makes sense.
2. In adding testings for changing the api user password, i hit issues where on a second run (after the node already had security installed) the templates and license would try to install using auth before the native realm had started. I've moved the templates upload after the native realm and inserted the pause for the native realm to come up (if configured) before trying to install the license and x-pack.  These should therefore be available by the time we try. I believe this is more robust and also simpler. @jakommo was familar with this

@jpcarey please check

jpcarey:
LGTM

PR #381, opened 2017-09-26
Allow to change directory name generation strategy so inventory_hostn…

pinkeen:
We hit a case on AWS that we lose all of the data when the instance is upgraded and the old storage attached to the new one. Upon running the provisioning the directory name changes.

Provides a fix for elastic/ansible-elasticsearch#384


Crazybus:
jenkins test this please

Issue #384, opened 2017-10-05
Issues when the inventory_hostname changes 


ServerNinja:
**Elasticsearch version**: 5.2.2

**Role version**:  `7714c925e06e8538a09b6e9d8c1bd074813e9639`

**JVM version** (`java -version`): `OpenJDK Runtime Environment (build 1.8.0_121-b13)`

**OS version** (`uname -a` if on a Unix-like system): (CentOS 7.3)
```Linux ip-192-168-0-145.common.bluescape.local 3.10.0-514.6.2.el7.x86_64 #1   SMP Thu Feb 23 03:04:39 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux```

**Description of the problem including expected versus actual behaviour**:

We had a situation where the inventory_name had changed from ansible's perspective, which caused the es_data_dirs to change the location of the data directory.

What happened:

For the affected node, the elasticsearch data was originally located in `/var/lib/elasticsearch/54.152.101.73-dev`, which was originally created by elasticsearch when the elasticsearch.yml.j2 template from this role populated the paths as such:

```
#################################### Paths ####################################

# Path to directory containing configuration (this file and logging.yml):
path.conf: /etc/elasticsearch/dev

path.data: /var/lib/elasticsearch/54.152.101.73-dev

path.logs: /opt/elasticsearch/logs/54.152.101.73-dev
```

When a change occurred with the elasticsearch instance in AWS which resulted in a different instance_hostname being reported from the AWS dynamic discovery, a new directory created after running the ansible playbook and elasticsearch was restarted. This new directory had been created and elasticsearch believed that it had lost all its shards on that node: `/var/lib/elasticsearch/192.168.0.145-dev`. 

This actually caused an outage where the elasticsearch cluster was in a "red" state until we discovered what had happened with the path.data change that the ansible-elasticsearch role had made in elasticsearch.yml. Fortunately, no data was lost.

In looking through the code to find out how this had happened, I saw that the `es_data_dirs` variable in ansible was being populated with information about the `inventory_hostname`, which seems to have caused this problem. 

https://github.com/elastic/ansible-elasticsearch/blob/7ab416c782f0a0d90430d1863150d405c1efdc3b/tasks/elasticsearch-parameters.yml#L51-L56

In the code from the link above, there is a comment on line 51 saying `#For directories we also use the {{inventory_hostname}}-{{ es_instance_name }} - this helps if we have a shared SAN.`

[...]

PR #386, opened 2017-10-05
Prefix data, pid, and log directories with cluster.name instead of inventory_hostname

ServerNinja:
This is to address the following issue:

https://github.com/elastic/ansible-elasticsearch/issues/384


Crazybus:
jenkins test this please

tasks/elasticsearch-parameters.yml
```diff
@@ -50,7 +50,7 @@
   when: use_system_d
 #For directories we also use the {{inventory_hostname}}-{{ es_instance_name }} - this helps if we have a shared SAN.
 
-- set_fact: instance_suffix={{inventory_hostname}}-{{ es_instance_name }}
+- set_fact: instance_suffix={{ inventory_hostname + '-' + es_instance_name if (es_shared_storage | default(false)) else es_config['cluster.name'] }}
```
Crazybus: This is [failing the build](https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/OS=ubuntu-1604,TEST_TYPE=standard,VERSION=5.x,label=swarm/29/console) because `es_shared_storage` isn't defined by default. I would recommend changing this to just `if es_shared_storage` and setting `es_shared_storage: False` in `defaults/main.yml`.

```
12:01:51        TASK [elasticsearch : set_fact] ************************************************
12:01:51        fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'cluster.name'\n\nThe error appears to have been in '/tmp/kitchen/elasticsearch/tasks/elasticsearch-parameters.yml': line 53, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact: instance_suffix={{ inventory_hostname + '-' + es_instance_name if (es_shared_storage | default(false)) else es_config['cluster.name'] }}\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n\nexception type: <class 'ansible.errors.AnsibleUndefinedVariable'>\nexception: 'dict object' has no attribute 'cluster.name'"}
```

PR #392, opened 2017-10-09
Parameterize timeout value for ES startup in Debian SysV init scripts

edge-records:
Simple, backwards-compatible change, to allow for other timeout values for node startup than the hard coded 10 seconds in Debian SysV init scripts. No other managed startup script/scheme seems to enforce a similar startup timeout, so only changed here.

Motivation: one of my 8 boxes that make up a cluster takes 11 seconds to startup, and the ansible run always fails due to this.


karmi:
Hi @edge-records, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/392.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

edge-records:
Sure, done

Crazybus:
Thanks for this @edge-records!

> Motivation: one of my 8 boxes that make up a cluster takes 11 seconds to startup, and the ansible run always fails due to this.

Ouch, that must have been a fun one to find. Thanks a lot for contributing this back though.

PR #403, opened 2017-11-28
Remove Java handling from role

### LINKED ISSUE #268: Proposal to Remove management of Java in Role
Currently we attempt to install java as well as refreshing certs - due to openjdk issues. Personally i feel our management of java is inadequate and there are superior roles for this. 

Either we enhance this component of the role or remove it entirely.  The former will add significant maintenance work and complexity for something that is not core to the roles purpose.  If we remove i would propose we check the environment and fail if it is not appropriately configured i.e. java8 installed for 5.x and certs working.

@jakommo @strootman @dliappis @barryib

### LINKED ISSUE #390: Java install fails - No package matching 'openjdk-8-jre-headless' is available
**Elasticsearch version**: 5.x

**Role version**:  master/7714c92

**JVM version**: 8

**OS version**: Ubuntu 14.04

When using role to install on Ubuntu 14.04, ansible fails with error: No package matching 'openjdk-8-jre-headless' is available

**Playbook**:

```yaml
- name: Simple Example
  hosts: vms
  remote_user: ---------
  roles:
    - role: elasticsearch
      es_instance_name: "node1"
  vars:
```

**Provide logs from Ansible**:

```
TASK [elasticsearch : Refresh java repo] ****************************************************************************************************************
ok: [---.---.---.---]

TASK [elasticsearch : Debian - Ensure Java is installed] ************************************************************************************************
fatal: [---.---.---.---]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'openjdk-8-jre-headless' is available"}
	to retry, use: -------------------/play.retry

PLAY RECAP **********************************************************************************************************************************************
---.---.---.---               : ok=16   changed=0    unreachable=0    failed=1   
```

janhoy:
Instead document how to use external role to install Java
Fixes #268, fixes elastic/ansible-elasticsearch#390


janhoy:
Anyone had a look at this yet?

shribigb:
@janhoy I think it misses following 2 key steps
1. Checks java is installed and the version is acceptable
2. Checks certs are valid

janhoy:
Can you elaborate on the certs check?

Crazybus:
@janhoy This is great stuff! Anything to simplify this role is a huge win!

Could you please rebase your branch against the latest master? Some additional changes will need to be made to the test-kitchen environment to make sure that a java role has been installed for testing.

janhoy:
@Crazybus I added a check that correct Java version is already installed.
Partly tested on Centos7. First run failed, after installing JAVA it succeeded. Have not tested the case where JAVA_HOME is explicitly set before running. The tasks should validate the java version of JAVA_PATH/bin/java first, then fallback to whatever java is on the path of ansible's "shell" command. 

Thinking about it, perhaps we should fail the role if JAVA_HOME is not explicitly set in the environment or alternatively set as some `es_java_home` variable in users playbook, since testing for java on ansible's path may is no guarantee that java is on the path when starting elastic.

janhoy:
My proposal is to introduce a new optional `es_java_home` variable. If not filled, system Java as found in JAVA_HOME or PATH will be used. If `es_java_home` is specified in playbook, then the role will validate that `es_java_home/bin/java` exists and is correct version, and `JAVA_HOME` will be set to that path in `elasticsearch.yml.j2`. Agree?

elasticcla:
Hi @janhoy, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/403.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

elasticcla:
Hi @janhoy, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/403.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

tgadiev:
Are there any plans to implement this change and remove Java handling?

jmlrt:
Hello,

- With Elasticsearch 7.x, java is no more installed by the role as it is embeded with elasticsearch package.
- With Elasticsearch 6.x, you are now able to set `es_java_install` variable to `false` to skip java installation.

PR #399, opened 2017-11-27
Allow for cluster-common config

ashokc:
Here is the driver for this change request

Most use cases for this role are in setting up a single cluster in a single run of the playbook. There are ES configurations that are shared across all host-groups & nodes in a cluster. For example the 'discovery hosts' can be the same across all nodes - master or data. In most use cases there is a single ES instance per host, so 9200 & 9300 are given and available. The config/scripts folder is kept in sync across the cluster nodes. So perhaps we can keep the playbook short & simple by separating out common ES configs from those that actually have to be different across host groups...    

We can allow the user to provide these common parameters under "group_vars/all.json" for example. In my case it looks like:

more group_vars/all.json

`{
  "private_iface" : "eth0",
  "public_iface" : "eth1",
  "masterHosts" : "{% for host in groups['es-master-nodes'] %} {{ hostvars[host]['ansible_'+public_iface]['ipv4']['address'] }}:9300 {% endfor %}"
,
  "es_scripts" : "false",
  "es_templates" : "false",
  "es_version_lock" : "false",
  "es_plugins" : [ {"plugin" : "ingest-geoip"} ],
  "common_es_config" : {"cluster.name": "es-cluster", "discovery.zen.ping.unicast.hosts" : "{{ masterHosts.split() }}", "http.port": 9200, "transp
ort.tcp.port": 9300, "bootstrap.memory_lock": false}
}`

In my case I have chosen to compute 'discovery hosts' from the known masterNodes group...

For example my inventory.yml looks like: (other users can use similar approaches that work for their inventory json/ini files)

es-master-nodes:
  hosts:

    es-master-1:
      ansible_host: 192.168.33.25
      ansible_user: vagrant
      ansible_ssh_private_key_file: .vagrant/machines/es-master-1/virtualbox/private_key

es-data-nodes:
  hosts:

    es-data-1:
      ansible_host: 192.168.33.26
      ansible_user: vagrant
      ansible_ssh_private_key_file: .vagrant/machines/es-data-1/virtualbox/private_key

    es-data-2:
      ansible_host: 192.168.33.27
      ansible_user: vagrant
      ansible_ssh_private_key_file: .vagrant/machines/es-data-2/virtualbox/private_key

PR #416, opened 2018-02-02
Cluster common configuration

ashokc:
1. Updated the file "templates/elasticsearch.yml.j2" for using cluster common configuration json snippet that can supplied for example via "group_vars/all.json"

2. Updated Readme.MD file with a sample usage for the same example as in the original repo.


Crazybus:
@ashokc Not sure what happened here but you somehow made a duplicate pull request. Closing this one since it is the same as #399

PR #421, opened 2018-02-08
6.2.0 compatible

neilime:


Crazybus:
jenkins test this please

Crazybus:
@neilime Thanks for this! I like the approach of symlinking the file however I'm not yet sure if this will properly when running multiple elasticsearch instances in a single host (there is an automated test for this). 

Also the [build is currently failing](https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/19/OS=ubuntu-1604,TEST_TYPE=standard,VERSION=6.x,label=swarm/console) when trying to create this symlink.

```
17:21:42        TASK [elasticsearch : Symlink Default Environment File] ************************
17:21:42        fatal: [localhost]: FAILED! => {"changed": false, "gid": 110, "group": "elasticsearch", "mode": "0660", "msg": "refusing to convert between file and link for /etc/default/elasticsearch", "owner": "root", "path": "/etc/default/elasticsearch", "size": 1613, "state": "file", "uid": 0}
```

Crazybus:
I ended up basing a new PR on this but instead making it an empty file instead of a symlink (due to the use case of running multiple elasticsearch instances on a single host as I explained above). https://github.com/elastic/ansible-elasticsearch/pull/431 . Thanks for finding the issue and helping out with the fix!

PR #423, opened 2018-02-13
Fix RedHat install - until condition was not correct

eRadical:
I was getting

`fatal: [redacted_hostname]: FAILED! => {"attempts": 5, "changed": false, "failed": true, "msg": "", "rc": 0, "results": ["elasticsearch-6.2.1-1.noarch providing elasticsearch is already installed"]}`

because „failed” was all the time present in the response.

### DIFF
--- a/tasks/elasticsearch-RedHat.yml
+++ b/tasks/elasticsearch-RedHat.yml
@@ -19,7 +19,7 @@
   when: es_use_repository
   register: redhat_elasticsearch_install_from_repo
   notify: restart elasticsearch
-  until: not redhat_elasticsearch_install_from_repo.failed
+  until: redhat_elasticsearch_install_from_repo.rc == 0
   retries: 5
   delay: 10
 
--- a/tasks/elasticsearch-RedHat.yml
+++ b/tasks/elasticsearch-RedHat.yml
@@ -19,7 +19,7 @@
   when: es_use_repository
   register: redhat_elasticsearch_install_from_repo
   notify: restart elasticsearch
-  until: '"failed" not in redhat_elasticsearch_install_from_repo'
+  until: not redhat_elasticsearch_install_from_repo.failed
   retries: 5
   delay: 10
 


Crazybus:
jenkins test this please

tasks/elasticsearch-RedHat.yml
```diff
@@ -19,7 +19,7 @@
   when: es_use_repository
   register: redhat_elasticsearch_install_from_repo
   notify: restart elasticsearch
-  until: '"failed" not in redhat_elasticsearch_install_from_repo'
+  until: not redhat_elasticsearch_install_from_repo.failed
```
Crazybus: Does this work for you? 

For me locally and the [jenkins job](https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/21/OS=centos-7,TEST_TYPE=standard,VERSION=6.x,label=swarm/console) fails consistently with

```
18:44:52        TASK [elasticsearch : RedHat - Install Elasticsearch] **************************
18:45:02        fatal: [localhost]: FAILED! => {"msg": "The conditional check 'not redhat_elasticsearch_install_from_repo.failed' failed. The error was: error while evaluating conditional (not redhat_elasticsearch_install_from_repo.failed): 'dict object' has no attribute 'failed'"}
``` 

It seems to be failing because the "failed" result doesn't exist until the task has finished. I just tried `until: redhat_elasticsearch_install_from_repo.rc == 1` and it seems to work as expected.



image

eRadical:
Indeed it does not work for new machines.

On the machine I initially tried it worked... but Ansible ran there many times.

Crazybus:
> Though I was wondering what is the rationale behind using until?
If the installation fails the first time and N-1 times why might it succeed in the N-th ?

Here is the commit where it was added https://github.com/elastic/ansible-elasticsearch/commit/e3c71a7f . 

My best guess would be that it was added to retry for any network connection errors. I can imagine it is pretty annoying if you are deploying a massive 100+ node cluster and a single node fails because of a network error with yum. 

> Can we just drop it?

Given the huge amount of different users and use-cases for this playbook I would rather leave it in unless there is a really compelling reason to remove it. I don't feel like this adds any extra complexity or maintenance to the playbook. Removing it could only cause some users to have connecting issues during yum installs. 


Were you able to test if `until: redhat_elasticsearch_install_from_repo.rc == 0` works properly for you?

eRadical:
Yes it is ok.
Tested on an already present machine and a new machine.

Crazybus:
Just fixed 6.2 support with https://github.com/elastic/ansible-elasticsearch/pull/431 If you rebase on master you should be able to test properly now.

Zauxst:
Got this issue when installing ElasticSearch-5.5.3 on CentOS 7, (+ had to also enable the repo "base") on the server for BitBucket Datacenter.

Fixed with until: `redhat_elasticsearch_install_from_repo.rc == 0`

Dagon-:
Hit this while installing 6.4.2 with ansible 2.5.2.
The workarounds already noted worked.

Crazybus:
jenkins test this please

Crazybus:
All tests are passing now!

Thanks for this @eRadical!

PR #425, opened 2018-02-15
Fix conditions in elasticsearch-security.yml

cyrilleverrier:
The condition was quoted and thus was a plain string.
Python interprets the string as always True
=> the condition was always True


pemontto:
Just saw this, I've done the same in #429 though there were a few more instances than those in elasticsearch-security.yml. Can either use #429 or update this PR 👍

cyrilleverrier:
Close this PR in favor of PR https://github.com/elastic/ansible-elasticsearch/pull/429. thanks @pemontto

PR #412, opened 2018-01-31
6.x testing -  Update tests to work with dynamic es_versions

Crazybus:

PR #429, opened 2018-02-16
Fix conditionals introduced in #412 and others

pemontto:
This change fixes failing conditionals that could be triggered by having xpack enabled but not having security in es_xpack_features.


Crazybus:
jenkins test this please

Crazybus:
@pemontto Thanks for this! Ansibles combination of yaml/jinja2/ansible always keeps it exciting 🎉

PR #433, opened 2018-03-05
auth fix by reordering set_fact on es_api_basic_auth_password

remil1000:
5.6.x with xpack security enabled: there may be some API auth issues when adding new members, also some idempotency issues on subsequent runs

With 6.x, the new method to setup password of elastic built-in user works perfectly;
but it looks like with 5.x, after initial setup there maybe some issues determining the correct password to use for API access.

This PR add a first try with the default 'changeme' password then switch to the native user password defined in es_users.native, this allows subsequent run with a non default password and also adding members after using the default creds before setting up native user/password

attached are the playbooks for 6.2.2 and 5.6.8 and an inventory file for a basic cluster

[playbook.5.6.8.yml.txt](https://github.com/elastic/ansible-elasticsearch/files/1781394/playbook.5.6.8.yml.txt)
[playbook.6.2.2.yml.txt](https://github.com/elastic/ansible-elasticsearch/files/1781396/playbook.6.2.2.yml.txt)
[inventory.txt](https://github.com/elastic/ansible-elasticsearch/files/1781397/inventory.txt)

### DIFF
--- a/tasks/xpack/security/elasticsearch-security-native.yml
+++ b/tasks/xpack/security/elasticsearch-security-native.yml
@@ -16,7 +16,31 @@
   set_fact: manage_native_roles=true
   when: es_roles is defined and es_roles.native is defined and es_roles.native.keys() | length > 0
 
+- name: set fact native_users
+  set_fact: native_users={{ es_users.native }}
+  when: manage_native_users
+
 #If the node has just has security installed it maybe either stopped or started 1. if stopped, we need to start to load native realms 2. if started, we need to restart to load
+
+# Try to list current users with the defined es_api_basic_auth_password
+- name: Try list Native Users
+  uri:
+    url: http://{{es_api_host}}:{{es_api_port}}/_xpack/security/user
+    method: GET
+    user: "{{es_api_basic_auth_username}}"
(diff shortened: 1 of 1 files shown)


Crazybus:
jenkins test this please

Crazybus:
@remil1000 Could you explain a bit more what issue you are running into exactly? There is actually already an [automated kitchen test](https://github.com/elastic/ansible-elasticsearch/blob/master/test/integration/xpack.yml) for this exact situation. The first run changes the `es_users.native.elastic` to `elasticChanged` and the second run has the `es_api_basic_auth_password` updated to `elasticChanged`. 

I would expect that if you update the password for the `es_users.native.elastic` user that you would then on the following run set `es_api_basic_auth_password` to be the new password. While your changes allows the user/password management to fallback to the password defined in `es_users.native` it still means that other functionality relying on `es_api_basic_auth_password` will still be broken such as the [Install templates with auth](https://github.com/elastic/ansible-elasticsearch/blob/eb82e1b055ddcac2ee9d9c1b38bcf9b093962a95/tasks/elasticsearch-template.yml#L34) task.

PR #434, opened 2018-03-08
Add a ES6 compatibility mode so that we can deploy ES6 configuration in ES 5.6 cluster

etiennecarriere:
I would like to have the possibility to validate the ES configuration on ES 5.6 before the migration to ES6. 
So I created a migration mode that can be activated on demand to have a configuration compliant to ES6 migration.


etiennecarriere:
Close this PR please. Was an error

PR #437, opened 2018-03-14
New STIG Xpack playbook

cchaffee:
The added files and playbook were created to meet a higher compliance for elasticsearch hardening inspec profile (https://github.com/elastic/elasticsearch-inspec). Primarily changes for audit settings, SSL, PKI, and permissions. Current compliance level at this pull request is 89.4%

Currently this works by importing the xpack.yml playbook and then building off of that. The intention was to keep the STIG playbook separate without making changes to current files in master. Newly created .kitchen.vagrant.yml should be used for running this with vagrant as the driver. Only CentOS supported at this time.


.kitchen.vagrant.yml
```diff
@@ -0,0 +1,101 @@
+---
+driver:
+  name: vagrant
```
Crazybus: Could you add this driver to the `Gemfile`depdencies.

.gitignore
```diff
@@ -8,3 +8,5 @@ TODO
 .idea/
 elasticsearch.iml
 !/vars/RedHat.yml
+.DS_store
```
Crazybus: I would recommend sticking this in your local [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). Much easier than having to add it to all projects.

.kitchen.vagrant.yml
```diff
@@ -0,0 +1,101 @@
...
+  attributes:
+    extra_vars:
+      es_major_version: "<%= ENV['VERSION'] %>"
+      <% if ENV['VERSION'] == '5.x' %>
+      es_version: '5.6.7'
+      <% end %>
+  <% end %>
+
+verifier:
+  name: inspec
```
Crazybus: Also needs to be added as a dependency to the Gemfile.

test/integration/xpack-stig.yml
```diff
@@ -0,0 +1,198 @@
+---
+# STIG based xpack provision for inspec hardening compliance
+# Make sure to change the variables below in the vars section
+# and in the kitchen.vagrant.yml file
+
+- import_playbook: roles/elasticsearch/test/integration/xpack.yml
```
Crazybus: This isn't working for me when I run it locally. The path is relative to this file so I think it needs to be something like `elasticsearch/test/integration/xpack.yml`

```
     ERROR! Unable to retrieve file contents
       Could not find or access '/tmp/kitchen/roles/elasticsearch/test/integration/xpack.yml'
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Converge failed on instance <xpack-stig-centos-7>.  Please see .kitchen/logs/xpack-stig-centos-7.log for more details
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

test/integration/xpack-stig.yml
```diff
@@ -0,0 +1,198 @@
...
+        - openssl
+        - rsyslog
+        - audit
+        - chrony
+        - ntp
+        - aide
+
+    - name: Copy instances.yml
+      template:
+        src: /tmp/kitchen/roles/elasticsearch/templates/instances.yml
```
Crazybus: Since this is only used for testing could you move the `instances.yml` into the `test/` directory.

.kitchen.vagrant.yml
```diff
@@ -0,0 +1,101 @@
...
+  output: ./%{platform}_%{suite}_results.json
+
+transport:
+  max_ssh_sessions: 6
+
+platforms:
+- name: centos-7
+  driver:
+    vagrantfiles:
+      - VagrantfileA.rb
```
Crazybus: Can you move this file into the `test` directory to keep the root directory clean.

.kitchen.vagrant.yml
```diff
@@ -0,0 +1,101 @@
...
+    name: vagrant
+    box: bento/centos-7
+    provision: true
+    pre_create_command: cp <%=ENV['ES_XPACK_LICENSE_FILE']%> {{vagrant_root}}
```
Crazybus: Is there a reason this can't be mounted as a volume the same way it is being done for the docker based tests? Mostly because `cp` won't be available for anyone testing from a windows host. 

```
      volume: 
        - <%=ENV['ES_XPACK_LICENSE_FILE']%>:/tmp/license.json
```

VagrantfileA.rb
```diff
@@ -0,0 +1,6 @@
+Vagrant.configure("2") do |config|
+  config.vm.provision "file", source: "license.json", destination: "/tmp/license.json"
+  config.vm.provision "shell", inline: "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py"
```
Crazybus: Is there a reason these can't be added as `provision_commands` in the kitchen file?

Example from the docker based tests.

```
  - name: debian-8
    driver_config:
      image: dliappis/debian-devopsci:8
      privileged: true
      provision_command:
        - apt-get update && apt-get -y install python python-dev python-pip build-essential libyaml-dev python-yaml curl wget
```

VagrantfileA.rb
```diff
@@ -0,0 +1,6 @@
+Vagrant.configure("2") do |config|
+  config.vm.provision "file", source: "license.json", destination: "/tmp/license.json"
```
Crazybus: This seems to be duplicated with the `pre_create_command: cp <%=ENV['ES_XPACK_LICENSE_FILE']%> {{vagrant_root}}` command. 

Is there a reason this can't be mounted as a volume the same way it is being done for the docker based tests?

```
      volume: 
        - <%=ENV['ES_XPACK_LICENSE_FILE']%>:/tmp/license.json
```

.kitchen.vagrant.yml
```diff
@@ -0,0 +1,101 @@
...
+      <% if ENV['VERSION'] == '5.x' %>
+      es_version: '5.6.7'
+      <% end %>
+  <% end %>
+
+verifier:
+  name: inspec
+  sudo: true
+  inspec_tests:
+    - git: https://github.com/elastic/elasticsearch-inspec.git
```
Crazybus: I wasn't able to get this far when testing locally. However how will this work? This is currently a private repo so my credentials will somehow need to be forwarded to inspec to be able to pull this repo.

.kitchen.vagrant.yml
```diff
@@ -0,0 +1,101 @@
...
+      branch: 6.0
```
Crazybus: Is there a way to pin this to a specific release/version? If this testing becomes part of the automated pull request testing I wouldn't want the tests to randomly fail just because the inspec tests had been updated.

.kitchen.vagrant.yml
cchaffee: Fixed

.kitchen.vagrant.yml
cchaffee: Fixed

.kitchen.vagrant.yml
cchaffee: We had some conversations on our end about that. We found that there were some significant changes between the 5.x and 6.x versions of elastic. At the time the inspec profile was built it was during 5.x so some controls were failing for me (I was on 6.1). 

The two camps of thought was either keep it one repo and make a branch for each major version or make separate specific repos for each version. I went with the branch way because I thought it would be "cleaner", but it can be changed too.

.kitchen.vagrant.yml
cchaffee: I'd have to look more into that I think. kitchen-docker and kitchen-vagrant differ in possible commands, I didn't find something similar to the kitchen-docker `volume:` one in kitchen-vagrant so I had to make do with the `pre_create_command:`.

VagrantfileA.rb
cchaffee: kitchen-vagrant doesn't support `provision_command:`. closest I could find was the `pre_create_command`

.kitchen.vagrant.yml
cchaffee: You should have access to that repo soon for the verifier part (having someone add you now). It will eventually be public as well.

cchaffee:
@crazybus thanks for all the comments! 

1. Yes, this is the plan. Reach out to Matt issett and Josh Bressers on your side - you are welcome to join our next sync/planning call

2. Yes, that's exactly what happened with `kitchen-docker`, some of the hardening settings don't work properly with a container (auditd was one I ran into, and other kernel level ones) so I switched over to `kitchen-vagrant`. The plan is to add some checks to those inspec tests so that if the virtualization method is a container, it doesn't apply. This way both virtualization platforms can be used.

3. After the fixes you should be able to do a kitchen create & converge on xpack-stig.yml. Let me know if something else happens when doing that? I had issues with `KITCHEN_YAML` environment variable working for me so I just linked `.kitchen.vagrant.yml` to `.kitchen.yml`

.kitchen.vagrant.yml
```diff
@@ -48,7 +48,7 @@ platforms:
 - name: centos-7
   driver:
     vagrantfiles:
-      - VagrantfileA.rb
+      - /test/Vagrantfile_setup.rb
```
Crazybus: Does this work for you? For me I need to remove the leading `/`. This would only work if you had this repo checked out in your root directory...I hope you aren't doing that :)

So for me it looks like `- test/Vagrantfile_setup.rb`

Gemfile
```diff
@@ -4,3 +4,6 @@ gem 'test-kitchen', '1.20.0'
 gem 'kitchen-docker', '2.6.0'
 gem 'kitchen-ansible', '0.48.1'
 gem 'net-ssh', '4.2.0'
+gem 'kitchen-vagrant', 	'1.3.0'
```
Crazybus: Thanks for adding these in, can you also checkin the updated `Gemfile.lock`?

test/integration/xpack-stig.yml
```diff
@@ -3,7 +3,7 @@
 # Make sure to change the variables below in the vars section
 # and in the kitchen.vagrant.yml file
 
-- import_playbook: roles/elasticsearch/test/integration/xpack.yml
+- import_playbook: elasticsearch/test/integration/xpack.yml
```
Crazybus: Hmm, still not working for me. 

```
 ERROR! the role 'elasticsearch' was not found in /tmp/kitchen/elasticsearch/test/integration/roles:/tmp/kitchen/roles:/tmp/kitchen/elasticsearch/test/integration

       The error appears to have been in '/tmp/kitchen/elasticsearch/test/integration/xpack.yml': line 7, column 7, but may
```

Is there anything else I need to setup to get this working? This is what I'm currently doing.

```
bundle install
export KITCHEN_YAML=.kitchen.vagrant.yml
export ES_XPACK_LICENSE_FILE="$(pwd)/license.json"
bundle exec kitchen converge xpack-stig-centos-7
```

.kitchen.vagrant.yml
Crazybus: I do actually already have access to this, I was mostly curious how this works from a test kitchen point of view. If it was using the git url (git@github.com:elastic/elasticsearch-inspec.git) then kitchen would be able to use my ssh agent to get the repo. However because it is using https it needs to have a GitHub access token configured somewhere. Is there some magic environment variable that needs to be exported so that test kitchen can use https auth somehow?

.kitchen.vagrant.yml
Crazybus: That makes a lot of sense. However I was actually referring to version of the elasticsearch-inspec project itself. If these tests become part of the automated testing pipeline for pull-requests I would want to have the version pinned to a specific git commit hash or version so that someone pushing changes to the elasticsearch-inspec repo wouldn't cause all pull requests to start failing in ansible-elasticsearch.

Crazybus:
@cchaffee Thanks for all the awesome work so far! Almost there!!! Thanks a lot for all of the explanations too about certain decisions that you made, all makes sense now :)

> Yes, this is the plan. Reach out to Matt issett and Josh Bressers on your side - you are welcome to join our next sync/planning call

Thanks, I have just pinged them for an invite!

> After the fixes you should be able to do a kitchen create & converge on xpack-stig.yml. Let me know if something else happens when doing that? I had issues with KITCHEN_YAML environment variable working for me so I just linked .kitchen.vagrant.yml to .kitchen.yml

Still running into some issues I'm afraid, I left some comments on the specific parts that are causing me errors right now.

test/integration/xpack-stig.yml
cchaffee: Hey @Crazybus, returning to this sorry about the long long delay. I got a new PC in between this time so I was trying to set this up fresh. It looks like what I ran into is that kitchen basically dumps the entire directry into `/tmp/kitchen` on the target machine. So either the directory `ansible-elasticsearch` needs to be renamed to just `elasticsearch` or the role needs to be changed to match the directory `ansible-elasticsearch`. Even if I try to just run any of the other xpack converges I run into that same error as you did. 

I just renamed the directory to elasticsearch and it starts to run the plays. Does that work for you? This way kitchen dumps the directory into the target as `elasticsearch` and then it can find it fine in `/tmp/kitchen/roles/elasticsearch` when looking at the `{ role: elasticsearch }`. I'm not even sure how I got around this now on my other machine.

Gemfile
cchaffee: Fixed

PR #461, opened 2018-06-26
add missing become: root

bilsch:
I'm not sure if there are other become statements missing but these were the ones I needed initially ;)

Are these known? I saw at least one in [460](https://github.com/elastic/ansible-elasticsearch/pull/460)

### DIFF
--- a/tasks/elasticsearch-config.yml
+++ b/tasks/elasticsearch-config.yml
@@ -93,13 +93,14 @@
     stat:
       path: "{{ sysd_script }}"
     register: sysd_stat_result
-  
+
   - name: Remove if it is a normal file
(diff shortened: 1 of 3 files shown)


karmi:
Hi @bilsch, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/461.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

itsmed:
jenkins test this please

Crazybus:
Jenkins test this please

PR #469, opened 2018-07-09
handle the nested response

leathekd:
The response from the search endpoint is now wrapped in a map with the
packages key.

I haven't updated DNS for the artifacts api, yet.  Aside from merging this after I update DNS, are there any other changes or steps needed here?


Crazybus:
@leathekd Thanks!

> are there any other changes or steps needed here?

Nope, this is perfect!

PR #481, opened 2018-08-04
Change include to import_tasks and include_tasks when suitable to get rid of deprecation warnings

Dumbaz:
Ansible deprecated the include statement to import tasks and will remove it entirely in the future.

This leads to the following deprecation warning:
```
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions. This feature will be
removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

Both keywords still work with tagging, so there should be no downside to using the newer terms everywhere.

I changed the import/include statements in every task, including the tests.

### DIFF
--- a/tasks/elasticsearch-Debian.yml
+++ b/tasks/elasticsearch-Debian.yml
@@ -65,7 +65,7 @@
     ES_PATH_CONF: "/etc/elasticsearch"
 
 - name: Debian - Include versionlock
-  include: elasticsearch-Debian-version-lock.yml
+  include_tasks: elasticsearch-Debian-version-lock.yml
   when: es_version_lock
 
 - name: Debian - Download elasticsearch from url
--- a/tasks/elasticsearch-RedHat.yml
+++ b/tasks/elasticsearch-RedHat.yml
@@ -10,7 +10,7 @@
   when: es_use_repository
 
 - name: RedHat - include versionlock
-  include: elasticsearch-RedHat-version-lock.yml
+  include_tasks: elasticsearch-RedHat-version-lock.yml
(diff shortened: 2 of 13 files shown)


Crazybus:
@Dumbaz Thanks for picking this up! Have ansible stated anywhere when `include` will be fully removed? I'm a bit confused because it is clearly marked as deprecated however their replacements are both marked as "preview".

> This module is flagged as preview which means that it is not guaranteed to have a backwards compatible interface.

Looking at [the changes from 2.4 to 2.5](https://docs.ansible.com/ansible/2.6/porting_guides/porting_guide_2.5.html#dynamic-includes-and-attribute-inheritance) they have already made breaking changes to how these modules work :(. Considering the "default" ansible version for this playbook (at least testing is 2.4) I'm not sure how I feel about introducing these changes until the features are moved from preview to something stable and backwards compatible. 

I'm not sure if the backwards compatibility changes this brings in are worth it to remove the deprecation warnings. I would hope that Ansible would only fully remove `include` once these modules are stable. 

I'd love to hear your thoughts on this though.

Dumbaz:
The feature was introduced in 2.4. The reason for implementing it was that the `include` statement basically did too much, When researching I also found a lot of issues at the Ansible repo with tags and some quirky behaviour with loops. So apparently the move to the new syntax is not as risk free as I thought.

I can offer to keep this PR up to date for the time being and try to get it all green until the modules become stable.

PR #487, opened 2018-08-17
add idempotence in check mode for plugins

lazouz:
this line : 
check_mode: no
ensures that the variable installed_plugins is defined, and the next task can be run in check mode ('undefined variable' instead)


lazouz:
hi, 

is anyone in charge of this ansible role ? I'm not sure what i propose is correct and if i follow the contrib guidelines

lazouz:
Hello,

Can I have a review ? 

cheers.

Crazybus:
jenkins test this please

lazouz:
thanks @Crazybus  :)

Crazybus:
@lazouz thanks for the fix!

Honestly I'm surprised that check mode works at all! It has always been so hit and miss for me in the past (referring to other ansible work).

PR #495, opened 2018-09-11
BACKPORTED: Fix RedHat install - until condition was not correct

nschlemm:
@eRadical thanks for https://github.com/elastic/ansible-elasticsearch/pull/423

### DIFF
--- a/tasks/elasticsearch-RedHat.yml
+++ b/tasks/elasticsearch-RedHat.yml
@@ -15,7 +15,7 @@
   yum: name=elasticsearch{% if es_version is defined and es_version != ""  %}-{{ es_version }}{% endif %} state=present update_cache=yes
   when: es_use_repository
   register: redhat_elasticsearch_install_from_repo
-  until: '"failed" not in redhat_elasticsearch_install_from_repo'
+  until: redhat_elasticsearch_install_from_repo.rc == 0
   retries: 5
   delay: 10
 


nschlemm:
@elasticmachine try again

and seriously consider using a more user and privacy friendly way than this bullshit CLA process you have currently!!!

xeraa:
> and seriously consider using a more user and privacy friendly way than this bullshit CLA process you have currently!!!

@nschlemm is the problem the concept of a CLA or something about our specific implementation?

nschlemm:
One can definitely discuss the concept of a CLA at length from different perspectives in the ambivalent space of open source businesses, but my objection concerns the pain your implementation (or more precisely that of the thirdparties you're using) entails by being not usable at all unless you allow a plethora of thirdparty requests just to submit a simple form that will send a confirmation link by email (...)

Not entirely sure how this plays out absolutely in terms of accessibility, but I sincerely invite you to experience that process using the (IMHO rightfully) popular https://github.com/gorhill/uMatrix/ in your browser - and all that for this tiny "contribution" of backporting an apparently long overdue issue thankfully already solved by the original contributor ;)

xeraa:
It is a valid concern and I can also see the issue with thirdparty cookies disabled. I've created an internal issue for the CLA checker, but I don't want to make any promises here. Especially since it's a one time thing it might not end up very high on our priority list. Though we should make contributions as painless as possible — let's see.

Anyway, back to the original issue, since this is a bit of a side quest here.

PR #510, opened 2018-11-13
* don't fetch APT key if existent

verboEse:
apt_key module of ansible does not support proxy settings (yet). To avoid errors for hosts that don't have a direct internet connection (or network traffic if the key is already existent), we should provide the key ID, so an admin could manually download the key (or with another ansible method) and ansible will not fail at this point.


elasticcla:
Hi @verboEse, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/510.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

tasks/elasticsearch-Debian.yml
```diff
@@ -14,7 +14,7 @@
 
 - name: Debian - Add Elasticsearch repository key
   become: yes
-  apt_key: url="{{ es_apt_key }}" state=present
+  apt_key: url="{{ es_apt_key }}" id=46095ACC8548582C1A2699A9D27D666CD88E42B4 state=present
```
Crazybus: Can you make this a variable please? Something like `es_apt_key_id: 46095ACC8548582C1A2699A9D27D666CD88E42B4` to make sure that it can be overriden for anyone who is overriding the `es_apt_key`.

verboEse: I just merged upstream into my branch and modified my changes accordingly. The Key-ID of the ES-APT-key now is configurable with the default set in var/Debian.yml

jmlrt:
jenkins test this please

jmlrt:
[oss-to-xpack-upgrade-ubuntu-1404](https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/75/OS=ubuntu-1404,TEST_TYPE=oss-to-xpack-upgrade,VERSION=7.x,label=linux/) is failing on CI but local test with `VERSION=7.x kitchen converge oss-to-xpack-upgrade-ubuntu-1404` is OK.

jmlrt:
jenkins test this please

PR #513, opened 2018-11-19
Update elasticsearch-parameters.yml

kakoni:
Fixing data_dirs name typo

### DIFF
--- a/tasks/elasticsearch-parameters.yml
+++ b/tasks/elasticsearch-parameters.yml
@@ -71,5 +71,5 @@
   set_fact: pid_dir={{ es_pid_dir }}/{{instance_suffix}}
 - name: set fact log_dir
   set_fact: log_dir={{ es_log_dir }}/{{instance_suffix}}
-- name: set fact log_dir
+- name: set fact data_dirs
   set_fact: data_dirs={{ es_data_dirs | append_to_list('/'+instance_suffix) }}


kakoni:
CLA signed. (karri.niemela@beans.fi)

Crazybus:
@kakoni It looks like you made the commits with a different email address. If you add the email you are using for these commits into your GitHub profile it will properly detect you as being verified. 

You can [see the email address here](https://github.com/elastic/ansible-elasticsearch/commit/a001c2ba1d0a917b8c89499a440abc9ad6c23f46.patch)

templates/elasticsearch.yml.j2
```diff
@@ -40,6 +40,10 @@ action.auto_create_index: {{ es_action_auto_create_index }}
 xpack.security.enabled: false
 {% endif %}
 
+{% if "monitoring" in es_xpack_features %}
```
Crazybus: As discussed in my reply in https://github.com/elastic/ansible-elasticsearch/issues/517#issuecomment-450179055 this goes against best practices to enable by default. Adding it down here would also make it impossible to disable via `es_config` because that variable is at the top of the file.

kakoni:
@Crazybus Fixed this so that it only includes the type fix.

Crazybus:
@kakoni the cla checker is still not happy I'm afraid :( 

Looking at the commits in https://patch-diff.githubusercontent.com/raw/elastic/ansible-elasticsearch/pull/513.patch it seems that you are committing with a different email address than the one you signed the cla with. If you add this email address (from the patch link) into your profile then it will work. Otherwise there is no way to match your commit messages up.

kakoni:
@Crazybus So;) One more try, changed commiter email so that its the same as with cla. (I do have two emails in profile also, but it seems that its not working?)

Crazybus:
> @Crazybus So;) One more try, changed commiter email for 69c49fc so that its the same as with cla. (I do have two emails in profile also, but it seems that its not working?)

@kakoni The problem is that you still have 3 commits in this PR all with different email addresses. You can see them listed in this URL: https://patch-diff.githubusercontent.com/raw/elastic/ansible-elasticsearch/pull/513.patch

kakoni:
Extra comments removed from PR, [patch](https://patch-diff.githubusercontent.com/raw/elastic/ansible-elasticsearch/pull/513.patch) looks decent.

Crazybus:
@kakoni woo! What a journey that was for such a small change. Thanks for finding and fixing the typo!

PR #516, opened 2018-11-27
Only attempt to copy the old users file if it actually exists

Crazybus:


Crazybus:
jenkins test this please

Issue #520, opened 2018-12-21
Is it possible pass from oss to basic license?


fedelemantuano:
I'd like to pass from `oss` to `basic` license. Is this possible without lose data?
If it's possible, maybe you should change the role, that can pass from basic to oss, but not vice versa.

**Elastic version**
6.5.1

**Playbook version**
6.5.1.1

**Java version**
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-b12)
OpenJDK 64-Bit Server VM (build 25.131-b12, mixed mode)

**OS version**
Linux xxxxx 3.10.0-693.el7.x86_64 #1 SMP Thu Jul 6 19:56:57 EDT 2017 x86_64 x86_64 x86_64 GNU/Linux
**Description of the problem including expected versus actual behaviour**:

**Playbook**:
```
#!/usr/bin/env ansible-playbook
---
- name: Install Elasticsearch Cluster
  hosts: elasticsearch
  become: true

  pre_tasks:
    - name: include commons variables
      include_vars: "vars/commons.yml"

  roles:
    - role: elasticsearch

  tasks:
    - name: Set vm.swappiness to 1
      sysctl:
        name: vm.swappiness
        value: 1
        state: present

  environment:
    http_proxy: "{{ http_proxy }}"
    https_proxy: "{{ https_proxy }}"
    no_proxy: "{{ no_proxy }}"
```

**Error**
```
Loaded plugins: langpacks, product-id, search-disabled-repos, subscription-
              : manager, versionlock
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Resolving Dependencies
--> Running transaction check
---> Package elasticsearch.noarch 0:6.5.4-1 will be installed
--> Processing Conflict: elasticsearch-oss-6.5.1-1.noarch conflicts elasticsearch
--> Restarting Dependency Resolution with new changes.
--> Running transaction check
---> Package elasticsearch-oss.noarch 0:6.5.1-1 will be updated
---> Package elasticsearch-oss.noarch 0:6.5.4-1 will be an update
--> Processing Conflict: elasticsearch-oss-6.5.4-1.noarch conflicts elasticsearch
--> Processing Conflict: elasticsearch-6.5.4-1.noarch conflicts elasticsearch-oss
--> Finished Dependency Resolution
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

[...]

PR #521, opened 2018-12-27
Allow switching between oss and standard packages

Crazybus:
Before only going from the standard package to oss was supported. Now
this works properly in both directions.

Fixes: #520

### DIFF
--- a/tasks/compatibility-variables.yml
+++ b/tasks/compatibility-variables.yml
@@ -14,6 +14,7 @@
     es_xpack_conf_subdir: ""
     es_repo_name: "{{ es_major_version }}"
     es_xpack_users_command: "elasticsearch-users"
+    es_other_package_name: "elasticsearch-oss"
 
 - name: Detect if es_version is before X-Pack was open and included
   set_fact:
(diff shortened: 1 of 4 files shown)


fedelemantuano:
Hi @Crazybus,

if you switch from `oss` to `xpack` or vice versa, you will have two [two yum repo](https://github.com/elastic/ansible-elasticsearch/blob/c2e51f8ecfdcb99c8e2b0b177b372ac7fcd432ee/tasks/elasticsearch-RedHat.yml#L14) in system.
To have a clean config it's better remove the old one.
I know that change only the name, maybe you could use a unique name for repos oss and xpack.

fedelemantuano:
Hi @Crazybus, I read the new commit but I don't understand [these lines](https://github.com/elastic/ansible-elasticsearch/blob/486717bc6f4936c244a6c9497f2f8686cab289dd/tasks/elasticsearch-RedHat.yml#L21).

Crazybus:
> Hi @Crazybus, I read the new commit but I don't understand these lines.

@fedelemantuano Good because it was wrong :P. I have fixed it up to remove the file instead of adding it (again). I also added some tests to make sure this is actually working properly. After adding the tests I found out that the Debian version was also only removing the package and not the repo so I have fixed that up too!

fedelemantuano:
Yes I saw the Debian part, so I'm fixing my [ansible-kibana](https://github.com/fedelemantuano/ansible-kibana) role too.

PR #523, opened 2018-12-27
Refactor deprecated version_compare filter calls

SlothOfAnarchy:
Previously, each of the calls generated a warning: "Using tests as filters is deprecated. Instead of using `result|version_compare` use `result is version_compare`. This feature will be removed in version 2.9.".

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -27,7 +27,7 @@ es_data_dirs: "/var/lib/elasticsearch"
 es_log_dir: "/var/log/elasticsearch"
 es_action_auto_create_index: true
 es_max_open_files: 65536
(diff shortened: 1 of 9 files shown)


elasticcla:
Hi @SlothOfAnarchy, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/523.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Crazybus:
jenkins test this please

Crazybus:
jenkins test this please

Crazybus:
Thanks for adding this in!

PR #526, opened 2019-01-14
Allow not installing Elasticsearch deb repository key

Fra-nk:
If a variable is set in Ansible, there is no way to unset it ever again, i.e. setting `es_apt_key: null` or `es_apt_key: ~` do not work.

Since this value is set in `defaults/main.yml` we have to perform a boolean check on the string.

### DIFF
--- a/tasks/elasticsearch-Debian.yml
+++ b/tasks/elasticsearch-Debian.yml
@@ -42,7 +42,7 @@
     apt_key:
       url: '{{ es_apt_key }}'
       state: present
-    when: es_apt_key is defined
+    when: es_apt_key | string
 
   - name: Debian - Add elasticsearch repository
     apt_repository:


Crazybus:
jenkins test this please

Crazybus:
```
10:44:56        TASK [elasticsearch : Debian - Add Elasticsearch repository key] ***************
10:44:56        fatal: [localhost]: FAILED! => {"msg": "The conditional check 'es_apt_key' failed. The error was: template error while templating string: expected token 'end of statement block', got '//'. String: {% if https://artifacts.elastic.co/GPG-KEY-elasticsearch %} True {% else %} False {% endif %}\n\nThe error appears to have been in '/tmp/kitchen/elasticsearch/tasks/elasticsearch-Debian.yml': line 41, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: Debian - Add Elasticsearch repository key\n    ^ here\n"}
```

Is the failure that CI is giving. Was this working for you when you tested it? Wondering if this could be a difference between ansible versions. In any case I think it is going to need to be checking for a non-empty string. 

Also the other GitHub status check is failing as you haven't signed the [CLA](https://www.elastic.co/contributor-agreement)

Fra-nk:
Ah, I only had tested with `es_apt_key: ""` since that's what we are using in our setup, but did not account for YAML string quoting shenanigans...

Validated it for the defaults now, too :).

> Also the other GitHub status check is failing as you haven't signed the CLA

Waiting for the my supervisor's approval to sign the CLA, but I've seen that, thanks :).

Crazybus:
jenkins test this please

Issue #525, opened 2019-01-11
'dict object' has no attribute 'stdout'


badfiles:
**Elasticsearch version**
elasticsearch-oss.noarch              6.5.4-1

**Role version**:  (If using master please specify github sha)
commit b37723aa1fabcd2ca5ce538995dd6e37253c1446

**JVM version** (`java -version`):
openjdk version "1.8.0_191"
OpenJDK Runtime Environment (build 1.8.0_191-b12)
OpenJDK 64-Bit Server VM (build 25.191-b12, mixed mode)

**OS version** (`uname -a` if on a Unix-like system):
ubuntu 18.04 (control), centos 7 (target)
**Description of the problem including expected versus actual behaviour**:
TASK [elastic.elasticsearch : correct java version selected] **************************************************************************************************
fatal: [host]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'stdout'\n\nThe error appears to have been in '/home/vlad/.ansible/roles/elastic.elasticsearch/tasks/java.yml': line 23, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: correct java version selected\n  ^ here\n"}

**Playbook**:
...
  roles:
    - dj-wasabi.zabbix-agent
    - geerlingguy.apache
    - geerlingguy.php
    - elastic.elasticsearch
...

**Provide logs from Ansible**:
ansible-2.7.5

TASK [elastic.elasticsearch : os-specific vars] ***************************************************************************************************************
task path: /home/vlad/.ansible/roles/elastic.elasticsearch/tasks/main.yml:2
ok: [*****] => {
    "ansible_facts": {
        "default_file": "/etc/sysconfig/elasticsearch",
        "es_home": "/usr/share/elasticsearch",
        "java": "{{ es_java | default('java-1.8.0-openjdk.x86_64') }}"
    },
    "ansible_included_var_files": [
        "/home/vlad/.ansible/roles/elastic.elasticsearch/vars/RedHat.yml"
    ],
    "changed": false
}

TASK [elastic.elasticsearch : Set the defaults here otherwise they can't be overriden in the same play if the role is called twice] ***************************
task path: /home/vlad/.ansible/roles/elastic.elasticsearch/tasks/compatibility-variables.yml:9
ok: [*****] => {
    "ansible_facts": {
        "es_install_xpack": false,
        "es_open_xpack": true,
[...]

PR #527, opened 2019-01-15
Execute java version check in check mode

katsukamaru:
This PR is related to this issue.
https://github.com/elastic/ansible-elasticsearch/issues/525

This problem is happened in --check mode, because in check mode ignored test for java_full_path and does not register it.
So this PR changed to execute in check mode and register java_full_path.

### DIFF
--- a/tasks/java.yml
+++ b/tasks/java.yml
@@ -18,6 +18,7 @@
   register: java_full_path
   failed_when: False
   changed_when: False
+  check_mode: no
   when: ansible_os_family == 'RedHat'
 
 - name: correct java version selected


Crazybus:
> So this PR changed to execute in check mode and register java_full_path.

Does this change allow you to run check-mode successfully? Your comment sounds like you meant to set `check_mode: yes` but the changes have it set to `no`.

Crazybus:
jenkins test this please

katsukamaru:
@Crazybus thank you for your message and added the results in my local environment. 

`check_mode: yes` means that ansible does not execute with --check.
My change `check_mode: yes` means actually execute in --check.

And below are results in my environment.

elastic/master branch
`ansible-playbook -i hosts playbook/es-c1.yml --check -v -l 172.20.0.54`
```
TASK [elastic.elasticsearch : Get the installed java path] ************************************************************************************************************************************************
skipping: [172.20.0.54] => {"changed": false, "msg": "skipped, running in check mode"}

TASK [elastic.elasticsearch : correct java version selected] **********************************************************************************************************************************************
fatal: [172.20.0.54]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'stdout'\n\nThe error appears to have been in '/home/admin/.ansible/roles/elastic.elasticsearch/tasks/java.yml': line 24, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: correct java version selected\n  ^ here\n"}

...
```

katsukamaru:java-version-check-in-check-mode branch

`ansible-playbook -i hosts playbook/es-c1.yml --check -v -l 172.20.0.54`
```
TASK [elastic.elasticsearch : Get the installed java path] ************************************************************************************************************************************************
ok: [172.20.0.54] => {"changed": false, "cmd": "update-alternatives --display java | grep '^/' | awk '{print $1}' | grep 1.8.0 | head -1", "delta": "0:00:00.021355", "end": "2019-01-15 17:57:41.289218", "failed_when_result": false, "rc": 0, "start": "2019-01-15 17:57:41.267863", "stderr": "", "stderr_lines": [], "stdout": "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.191.b12-0.amzn2.x86_64/jre/bin/java", "stdout_lines": ["/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.191.b12-0.amzn2.x86_64/jre/bin/java"]}

TASK [elastic.elasticsearch : correct java version selected] **********************************************************************************************************************************************
ok: [172.20.0.54] => {"changed": false}

...
```

In katsukamaru:java-version-check-in-check-mode branch actually executes the command and returns ok.
Please let me know if there are any problems.

Crazybus:
> check_mode: yes means that ansible does not execute with --check. 
> My change check_mode: yes means actually execute in --check.

I think you meant `check_mode: no` here right?

I just had a look at the ansible [check mode](https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html#id2) docs and it looks like I got it around backwards. 

> Force a task to run in check mode, even when the playbook is called without --check. This is called check_mode: yes.
Force a task to run in normal mode and make changes to the system, even when the playbook is called with --check. This is called check_mode: no.

Pretty confusing actually, but it means that your change does look good then! :)

katsukamaru:
@Crazybus
> I think you meant `check_mode: no` here right?
You are right. I fixed my comment `check_mode: yes` to `check_mode: no`. Thank you for pointing it out.

I signed the CLA and seem to be passed all checks.

Crazybus:
Thanks for your contribution @katsukamaru and teaching me how the check_mode actually works :D

PR #528, opened 2019-01-15
Use systemd's RequiresMountsFor

Fra-nk:
User might be interested in separating ephemeral and persistent data.
Among others, this can be done by putting `data_dir` on another disk.

When doing that in cloud environments via [`cloud-init`](https://github.com/cloud-init/cloud-init) or [`waagent`](https://github.com/Azure/WALinuxAgent) you need to time the service start to _after_ the necessary devices are mounted, otherwise a low run-level can result in trying to start the service before all data/mounts is/are available.

This PR adds [systemd's `RequiresMountsFor`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#RequiresMountsFor=) directive for the corresponding folders.

### DIFF
--- a/templates/systemd/elasticsearch.j2
+++ b/templates/systemd/elasticsearch.j2
@@ -3,6 +3,8 @@ Description=Elasticsearch-{{es_instance_name}}
 Documentation=http://www.elastic.co
 Wants=network-online.target
 After=network-online.target
+{# Directive 'WorkingDirectory' creates an implicit dependecy for {{es_home}}, so it can be omitted here #}
+RequiresMountsFor={{ data_dirs | array_to_str(separator=' ') }} {{log_dir}} {{pid_dir}} {{conf_dir}}
 
 [Service]
 Environment=ES_HOME={{es_home}}


Crazybus:
jenkins test this please

templates/systemd/elasticsearch.j2
```diff
@@ -3,6 +3,8 @@ Description=Elasticsearch-{{es_instance_name}}
 Documentation=http://www.elastic.co
 Wants=network-online.target
 After=network-online.target
+{# Directive 'WorkingDirectory' creates an implicit dependecy and can be omitted #}
```
Crazybus: Did you mean to leave this in? This comment mentions the `WorkingDirectory` directive but you are adding in `RequiresMountsFor`.

Fra-nk: The `WorkingDirectory` directive is used in [line 18](https://github.com/elastic/ansible-elasticsearch/blob/e999988560469d69d353ffb472c7f83ab685dccd/templates/systemd/elasticsearch.j2#L18).
It **implicitly** creates the necessary `After=`/`Requires=` directives for `{{es_home}}`, therefore we can skip adding `{{es_home}}` to the list of folders in `RequiresMountsFor`.

I'll refine the comment to make it more clear, why I skipped `{{es_home}}`.

Do you prefer Jinja2 comments here, i.e. comments that are no longer in the templated file, or "regular" comments?

Fra-nk: Updated the Jinja comment for now - since we're using template variables, it would imho be confusing in the final systemd-unitfile.

Fra-nk:
I would consider the changes as resolved, until further notice.

Waiting for feedback :)

templates/systemd/elasticsearch.j2
Crazybus: Thank you for clarifying and updating it!

Crazybus:
jenkins test this please

Crazybus:
Thank you @Fra-nk for adding this in! 

(And for teaching me about a cool systemd feature)

PR #539, opened 2019-02-28
make playbook (tested for debians) compatible with the  mode

grzegorznowak:
Hey guys,

it was impossible to run the `--check` mode on top of the role.
(Please refer to the Ansible docs for details at: https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html)

The PR adds a micro tweak to properly `register` variables when in the check_mode.

### DIFF
--- a/tasks/elasticsearch-Debian.yml
+++ b/tasks/elasticsearch-Debian.yml
@@ -15,6 +15,7 @@
     register: elasticsearch_package
     failed_when: False
     changed_when: False
+    check_mode: no
 
   - name: stop elasticsearch
     service:
--- a/tasks/java.yml
+++ b/tasks/java.yml
@@ -44,6 +44,7 @@
   register: open_jdk
   ignore_errors: yes
   changed_when: false
+  check_mode: no
 
 #https://github.com/docker-library/openjdk/issues/19 - ensures tests pass due to java 8 broken certs
 - name: refresh the java ca-certificates


Crazybus:
jenkins test this please

grzegorznowak:
thanks @Crazybus , over the course of time I've found at least one more instance where the `check mode` was failing due to the same reason. 
I think it would make no difference to any of the role's properties/working if we just added the same `check_mode: no` attribute to any instruction that `registers` a variable.
Let me know if you think that's not a good idea, otherwise I can issue PR that mass-changes that in one go throughout.

Crazybus:
> I think it would make no difference to any of the role's properties/working if we just added the same check_mode: no attribute to any instruction that registers a variable. Let me know if you think that's not a good idea

It sounds sane to me as long as the shell command is not modifying state. 

> otherwise I can issue PR that mass-changes that in one go throughout.

Go for it! I'll review it extra carefully :)

grzegorznowak:
ok that is now done in https://github.com/elastic/ansible-elasticsearch/pull/542 @Crazybus .
Not too much of what could be added, and as per my comment in that PR, I think for this role to be up to the current devops expectations, it has to get rid of all the `registers` anyway.
From what I remember, there are some easy ones that can be redone into `notify` handlers. And the rest probably could use a custom module or other means if necessary.

Issue #330, opened 2017-07-05
Ansible dry run (--check) fails due to missing variables


sangdrax8:
Running a "dry run" with ansible fails because some variables are not populated in check mode making other tasks error.  This can be fixed by setting those commands which populate varaibles used by other tasks (and are them self safe to always run) with check_mode: no.  The following diff's are what I needed to fix in my install, but there could be others like them in sections that I am not using

```
diff --git a/tasks/elasticsearch-plugins.yml b/tasks/elasticsearch-plugins.yml
index 8ec9e25..7d764af 100644
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -16,6 +16,7 @@
   register: installed_plugins
   changed_when: False
   ignore_errors: yes
+  check_mode: no
   environment:
     CONF_DIR: "{{ conf_dir }}"
     ES_INCLUDE: "{{ instance_default_file }}"
diff --git a/tasks/java.yml b/tasks/java.yml
index c0c63b8..990899a 100644
--- a/tasks/java.yml
+++ b/tasks/java.yml
@@ -21,6 +21,7 @@
 - shell: java -version 2>&1 | grep OpenJDK
   register: open_jdk
   ignore_errors: yes
+  check_mode: no
   changed_when: false
```

gingerwizard:
@sangdrax8 i'm trying to ensure the role works with `--check` .  Some tasks generate variables which are subsequently used, yet they perform state changing actions. For example,

https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-Debian.yml#L27

Fails in dry run mode if the elasticsearch repo is not added in the previous task. I can set this previous tasks to dry_run but this seems to somewhat defeat the purpose.

Furthermore, 

https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-plugins.yml#L50

fails as it checks its own output via` plugin_installed.rc`.  This command can't be set to `check_mode:no` as it makes state changing actions are relies on ES being instanced. Where you able to overcome these issues?

gingerwizard:
I think we can aspire to supporting check mode on the 2nd run but im not convinced its going to be possible on the 1st run.

gingerwizard:
[...]

PR #540, opened 2019-03-07
Additional support for check mode on Debian

andyshinn:
This is related to #487, #377, #330, and and #527. The latest PR in #527 helped get check mode working but seemed only for the RedHat variant. This gets check mode working on an already installed Elasticsearch Debian node. The checks appear to just be listing commands so should be harmless in check mode.


Crazybus:
This one is an exact duplicate of #539. The other one was opened first so I will close this one. You were just too late by a few days :P

Issue #538, opened 2019-02-15
Installation via ansible-galaxy does not work with ansible-galaxy 2.7.6 when omitting the version


NikolaiGulatz:
**Ansible version**
`2.7.6`

**Ansible galaxy version**
`2.7.6`

**Role version**:
master (`0ccf60831e0d3d38ca14b67bcb1e316acab336ae`)

**OS version** (`uname -a` if on a Unix-like system):
Arch

**Description of the problem including expected versus actual behaviour**:
When I run `ansible-galaxy install elastic.elasticsearch` I get the following warning:

```
- downloading role 'elasticsearch', owned by elastic
 [WARNING]: - elastic.elasticsearch was NOT installed successfully: Unable to compare role versions (0.1.0, 0.1.2, 5.5.1,
6.1.3, 6.2.2, 6.2.3, 6.2.4, 6.3.0, 6.3.0.1, 6.4.0, 6.5.1, 6.6.0, 6.3.0-1, 6.3.1, 0.1.1, 6.2.4-1) to determine the most
recent version due to incompatible version formats. Please contact the role author to resolve versioning conflicts, or
specify an explicit role version to install.
```

When specifying a version it works as expected.

yujinov:
You can install specific version e.g. `ansible-galaxy install elastic.elasticsearch,6.6.0`

```
$ ansible-galaxy install elastic.elasticsearch
- downloading role 'elasticsearch', owned by elastic
 [WARNING]: - elastic.elasticsearch was NOT installed successfully: Unable to compare role versions (0.1.0, 0.1.2, 5.5.1, 6.1.3, 6.2.2, 6.2.3, 6.2.4,
6.3.0, 6.3.0.1, 6.4.0, 6.5.1, 6.6.0, 6.3.0-1, 6.3.1, 0.1.1, 6.2.4-1) to determine the most recent version due to incompatible version formats. Please
contact the role author to resolve versioning conflicts, or specify an explicit role version to install.

$ ansible-galaxy install elastic.elasticsearch,6.6.0
- downloading role 'elasticsearch', owned by elastic
- downloading role from https://github.com/elastic/ansible-elasticsearch/archive/6.6.0.tar.gz
- extracting elastic.elasticsearch to /Users/user/.ansible/roles/elastic.elasticsearch
- elastic.elasticsearch (6.6.0) was installed successfully
```

Crazybus:
> Please contact the role author to resolve versioning conflicts, or
[...]

PR #541, opened 2019-03-11
Update readme to include versioned install command

Crazybus:
In #538 it was discovered that recent versions of the ansible galaxy
command will fail if there aren't semver compatible releases in the
history.

The error says:

> Please contact the role author to resolve versioning conflicts, or
> specify an explicit role version to install.

Removing the old releases is not such a great idea so instead updating
the install command is the best option.

Fixes: #538

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ This role uses the json_query filter which [requires jmespath](https://github.co
 Create your Ansible playbook with your own tasks, and include the role elasticsearch. You will have to have this repository accessible within the context of playbook.
 
 ```sh
-ansible-galaxy install elastic.elasticsearch
+ansible-galaxy install elastic.elasticsearch,6.6.0
 ```
 
 Then create your playbook yaml adding the role elasticsearch. By default, the user is only required to specify a unique es_instance_name per role application.  This should be unique per node. 


grzegorznowak:
hey @Crazybus , I understand the concern you're having here, but remember that any extra step devs have to take, makes a project less likely to be used. You are already way down in the usage on galaxy, despite hosting the "official" Elasticsearch role. I would strongly suggest fixing the release tags on github and importing back into galaxy.
Have faced the same issue with my own roles, and if I remember correctly it's all about updating releases git tags' names in accordance to the version pattern galaxy is using: `x.y.z`, no need to rename any release per-se.
Hopefully that can be done without too much pita, as honestly, I'm too reluctant to using a role that needs it's version typed-in in order to install the latest.

Crazybus:
> Have faced the same issue with my own roles, and if I remember correctly it's all about updating releases git tags' names in accordance to the version pattern galaxy is using: x.y.z, no need to rename any release per-se.

Got any links to documentation on how this works? It doesn't sound right that the git tags don't need to be matching the release tags. If there really is a way to fix up the release tags without deleting any releases them I'm totally willing to do it. However deleting releases is not an option as users rely on these existing to deploy Elasticsearch. 

>  Hopefully that can be done without too much pita, as honestly, I'm too reluctant to using a role that needs it's version typed-in in order to install the latest.

I mean this in the nicest way possible, but if needing to pin a version is too much effort then you probably don't have enough time to setup and maintain an Elasticsearch cluster. I would recommend that you use a hosted service like [Elastic Cloud](https://www.elastic.co/cloud/) instead. I can't force you, but I would also strongly recommend against installing the latest without specifying a version. Doing this gives you no guarantees around making sure that yourself/others/CI systems are all using the same version of the role.

grzegorznowak:
>  if needing to pin a version is too much effort then you probably don't have enough time to setup and maintain an Elasticsearch cluster

please let's not go down this route @Crazybus, I maintain enough ES and non-ES machines to know where to put my effort best ;)

So the actual use-case as I see it with having to manually specify the tag is inability to put self-building automation to work, without having to have a bespoke routine that updates the tag via querying github or other means of current-truth verification. I can't nor I em willing to build and maintain it at this stage.

And let's be clear, I'm all about freezing versions on production servers, but automated integration builds are there for purpose, and this is being able to say with highest possible confidence, that my cluster will integrate with latest version just fine.  You know, for the purpose of quickly going from version to version in case a zero-day CVE gets reported etc.

grzegorznowak:
> Got any links to documentation on how this works?
no documentation, sorry. I've done it once a month or so ago, exactly for the same purpose.
In the end I've deleted and re-issued all the releases that had wrong tag assigned, but with a proper naming convention this time. I think! From what I see it's only 2 that are offending for this repo.

It's not very fortunate of the Galaxy to be having to do that, I agree, and possibly they should be having a way to expose versions manually via `meta.yml`, which I couldn't find they do however, but in the end this will be one off change. And even if breaks someone's installation, they will know better to double check on the project's README page where I'd put a big fat notifications about tags' change. 
That's what I'd do anyway!

Crazybus:
Ignoring the whole discussion, is installing it without version not giving you the latest btw? It seems like whatever Ansible galaxy did that broke getting the latest without a version is now fixed. 

```
❯ ansible-galaxy install elastic.elasticsearch
- downloading role 'elasticsearch', owned by elastic
- downloading role from https://github.com/elastic/ansible-elasticsearch/archive/7.1.1.tar.gz
- extracting elastic.elasticsearch to /Users/mick/.ansible/roles/elastic.elasticsearch
- elastic.elasticsearch (7.1.1) was installed successfully
```

---

> no documentation, sorry. I've done it once a month or so ago, exactly for the same purpose.
In the end I've deleted and re-issued all the releases that had wrong tag assigned, but with a proper naming convention this time. I think! From what I see it's only 2 that are offending for this repo.

If you have to delete releases then I'm afraid that's not an option. We once removed a bunch of really old releases (that were hosted on a 5 year old artifact server) and saw a massive amount of fallout from users that were still using them. As you said earlier, this is the official role from Elastic, I can't get away with just removing old releases that users are using.

grzegorznowak:
yes, hm, let's compare versions then, as I'm seeing this instead:
```
ansible-galaxy install elastic.elasticsearch --force
- changing role elastic.elasticsearch from 6.3.1 to unspecified
- downloading role 'elasticsearch', owned by elastic
 [WARNING]: - elastic.elasticsearch was NOT installed successfully: Unable to compare role versions (0.1.0, 0.1.2, 5.5.1, 6.1.3, 6.2.2, 6.2.3, 6.3.0, 6.3.0.1, 6.4.0, 6.5.1, 6.6.0, 6.3.0-1, 6.3.1, 0.1.1, 6.2.4-1, 7.1.1, 6.2.4) to determine the most recent version due to
incompatible version formats. Please contact the role author to resolve versioning conflicts, or specify an explicit role version to install.
```

```
ansible-galaxy 2.8.2
```

Crazybus:
Also 2.8.2:

```
❯ ansible-galaxy --version && ansible-galaxy install elastic.elasticsearch
ansible-galaxy 2.8.2
- downloading role 'elasticsearch', owned by elastic
- downloading role from https://github.com/elastic/ansible-elasticsearch/archive/7.1.1.tar.gz
- extracting elastic.elasticsearch to /Users/mick/.ansible/roles/elastic.elasticsearch
- elastic.elasticsearch (7.1.1) was installed successfully
```

You could also install from master using `ansible-galaxy install git+https://github.com/elastic/ansible-elasticsearch`. If your goal is to catch breakages early this will allow you to see them once they hit master instead of waiting for each release.

grzegorznowak:
hm, seem like there must be some config discrepancy between our ansible-galaxy configs somewhere behind the lines @Crazybus . I'm pretty sure mine is as default as possible, but let's not spin wheels on trying to debug that. Must be an issue elsewhere. Thanks for trying though, and feel free to ping me here if you wanted to check something else around that specifically.

I could install from master yes, but what I really want to test is if the latest *release* passes as a part of my own set of playbooks, in various production-ready configurations. I'm not going to be using master directly for deploys ever. Not by design, that is. 
Cheers

PR #542, opened 2019-03-14
add a few more 'check_mode:no' props for compatiblity with --check

grzegorznowak:
Those look ok to add the check_mode:no flag, since not changing anything in the system.
But ultimately all the `register:` should probably just go this way or another I'm afraid, for ultimate idempotence of the role.


Crazybus:
jenkins test this please

PR #548, opened 2019-03-21
[ci] remove vault url from defaults

jonahbull:
We moved this default inside JJBB itself so it's not necessary to have
it in the defaults file anymore.


jonahbull:
@Conky5 beat me to it: elastic/ansible-elasticsearch@18c9332

PR #550, opened 2019-03-28
Fix template conditional

pemontto:
The conditional for unauthenticated uploading of templates would always run regardless of whether the templates files had changed or not.

### DIFF
--- a/tasks/compatibility-variables.yml
+++ b/tasks/compatibility-variables.yml
@@ -23,7 +23,7 @@
     es_open_xpack: false
   when: "es_version is version_compare('6.3.0', '<')"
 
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @pemontto, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/550.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

pemontto:
Email issue fixed

Crazybus:
jenkins test this please

PR #555, opened 2019-04-16
add ability to set processors count in elasticsearch.yml

kcm:
Add ability to set `processors` in `elasticsearch.yml` per:

https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-threadpool.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,7 @@ In addition to es_config, the following parameters allow the customization of th
       - plugin: ingest-geoip 
   ```
 * ```es_path_repo``` Sets the whitelist for allowing local back-up repositories
+* ```es_processors``` (e.g. `16`) Sets the number of processors (cores). Used in sizing thread pools. Useful for multitenancy or when the value is not correctly detected.
 * ```es_action_auto_create_index ``` Sets the value for auto index creation, use the syntax below for specifying indexes (else true/false):
      es_action_auto_create_index: '[".watches", ".triggered_watches", ".watcher-history-*"]'
 * ```es_allow_downgrades``` For development purposes only. (true or false (default) )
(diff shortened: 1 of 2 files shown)


kcm:
Looking at build failures in https://github.com/elastic/infra/issues/10875

kcm:
@Crazybus can you retrigger the CI job? Closing/reopening trick doesn't work anymore without write access.

Crazybus:
jenkins test this please is the phrase (this should work for you too)

kcm:
jenkins test this please

templates/elasticsearch.yml.j2
```diff
@@ -35,6 +35,10 @@ action.auto_create_index: false
 action.auto_create_index: {{ es_action_auto_create_index }}
 {% endif %}
 
+{% if es_processors is defined %}
```
Crazybus: What is the reason for not just specifying this in the `es_config` variable? Talking about doing it like this:

```
    es_config:
      processors: 16
```

README.md
```diff
@@ -399,6 +399,7 @@ In addition to es_config, the following parameters allow the customization of th
       - plugin: ingest-geoip 
   ```
 * ```es_path_repo``` Sets the whitelist for allowing local back-up repositories
+* ```es_processors``` (e.g. `16`) Sets the number of processors (cores). Used in sizing thread pools. Useful for multitenancy or when the value is not correctly detected.
```
Crazybus: We are planning to drop multi-tenancy support as part of the 7.x release. Given the warnings in https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-threadpool.html#processors about this being an expert level use-case I'm a bit hesitant to include a specific setting just for this. Especially when it can already be set directly in `es_config`

kcm:
It does fit better directly now that I've taken a closer look at the project, thanks - closing.

PR #556, opened 2019-04-26
[test] fix debian-8 provisioning in kitchen tests

jmlrt:
Jessie backport repos have been removed from debian mirrors and moved to archive mirrors (cf. https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html). We also need to disable validity checks as this repo is no more updated (https://unix.stackexchange.com/questions/508724/failed-to-fetch-jessie-backports-repository)

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -76,7 +76,7 @@ platforms:
       privileged: true
       provision_command:
         - echo "deb http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
-        - echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.con
+        - echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf
         - apt-get update && apt-get -y install -t jessie-backports openjdk-8-jre-headless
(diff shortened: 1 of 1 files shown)


jmlrt:
this should fix https://github.com/elastic/infra/issues/11013

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

.kitchen.yml
```diff
@@ -75,7 +75,8 @@ platforms:
       image: debian:8
       privileged: true
       provision_command:
-        - echo "deb http://http.debian.net/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
+        - echo "deb http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
+        - echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.con
```
Crazybus: Should this be `/etc/apt/apt.conf`?

jmlrt: fixed in https://github.com/elastic/ansible-elasticsearch/pull/556/commits/7488a569ac9925778f875e803ea8e96e36982d2b

PR #558, opened 2019-05-02
[7.x] add support for elasticsearch 7.x

jmlrt:


jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

tasks/elasticsearch-parameters.yml
```diff
@@ -13,12 +13,12 @@
   when: es_config['http.port'] is not defined
 
 - name: debug message
-  debug: msg="WARNING - It is recommended you specify the parameter 'transport.tcp.port'"
-  when: es_config['transport.tcp.port'] is not defined
+  debug: msg="WARNING - It is recommended you specify the parameter 'transport.port'"
+  when: es_config['transport.port'] is not defined
```
Crazybus: This will be a bit confusing for someone using 6.x since it will say to set an option that is not available. 

I think it would be better to just remove this debug line altogether since it is already mentioned in the readme.

tasks/elasticsearch-parameters.yml
```diff
@@ -13,12 +13,12 @@
...
 
 - name: debug message
-  debug: msg="WARNING - It is recommended you specify the parameter 'discovery.zen.ping.unicast.hosts'"
-  when: es_config['discovery.zen.ping.unicast.hosts'] is not defined
+  debug: msg="WARNING - It is recommended you specify the parameter 'discovery.seed_hosts'"
+  when: es_config['discovery.seed_hosts'] is not defined
```
Crazybus: Same thing goes for this as I said for the `es_config['transport.port']` debug message.

tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+
+- set_fact: "es_major_version={{ es_version[0] }}.x"
```
Crazybus: This is gonna break when version 10 comes out ;).

Maybe it would be nicer to instead do something like `{{ es_version.split('.')[0] }}`

test/integration/helpers/serverspec/multi_spec.rb
```diff
@@ -11,16 +11,11 @@
   describe file("/etc/elasticsearch/#{vars['es_instance_name']}/elasticsearch.yml") do
     it { should be_file }
     it { should contain 'http.port: 9201' }
-    it { should contain 'transport.tcp.port: 9301' }
```
Crazybus: I think this should now be `it { should contain 'transport.port: 9301' }` instead of being removed right?

test/integration/multi.yml
```diff
@@ -11,9 +11,7 @@
     es_data_dirs:
       - "/opt/elasticsearch/master"
     es_config:
-      discovery.zen.ping.unicast.hosts: "localhost:9300"
       http.port: 9200
-      transport.tcp.port: 9300
```
Crazybus: I think this should still be left in as `transport.port` otherwise it goes against our recommendation of always setting it.

test/integration/xpack-upgrade.yml
```diff
@@ -8,21 +8,26 @@
...
-      xpack.security.authc.realms.native1.type: "native"
+      xpack.security.authc.realms.file1.type: file
       xpack.security.authc.realms.native1.order: 1
+      xpack.security.authc.realms.native1.type: native
+    es_config_7x:
+      http.port: 9200
+      xpack.security.enabled: True
+      xpack.security.authc.realms.file.file1.order: 0
+      xpack.security.authc.realms.native.native1.order: 1
+    es_config: "{{ es_config_7x if es_major_version == '7.x' else es_config_6x }}"
```
Crazybus: I like this trick a lot :)

tasks/main.yml
```diff
@@ -1,4 +1,11 @@
...
+  when:
+      - not es_major_version
```
Crazybus: This needs to be something like `es_major_version not defined`

Otherwise you get this error:

```
     TASK [elasticsearch : set_fact] ************************************************
       fatal: [localhost]: FAILED! => {"msg": "The conditional check 'not es_major_version' failed. The error was: error while evaluating conditional (not es_major_version): 'es_major_version' is undefined\n\nThe error appears to have been in '/tmp/kitchen/elasticsearch/tasks/main.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact: \"es_major_version={{ es_version[0] }}.x\"\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}
```

test/integration/multi.yml
jmlrt: well the goal was to avoid managing different config for 6.x and 7.x like that:
```

    es_config_6x:
      http.port: 9200
      transport.tcp.port: 9300
      node.data: false
      node.master: true
      bootstrap.memory_lock: tru
    es_config_7x:
      http.port: 9200
      transport.tcp.port: 9300
      node.data: false
      node.master: true
      bootstrap.memory_lock: tru
    es_config: "{{ es_config_7x if es_major_version == '7.x' else es_config_6x }}"
```

But we can add it

PR #560, opened 2019-05-10
[xpack] use elasticsearch default xpack features

jmlrt:
- Stop forcing es_xpack_features variable in order to let elasticsearch install default features described in http://localhost:9200/_xpack
- Change xpack test scope to be able to test default xpack install
  - xpack scenario will test xpack install with default features
  - xpack upgrade scenario will fully test security feature
  - oss-to-xpack-upgrade will test installing only other specific features
- Cleanup some duplicate serverspec tests
- Remove `system_key`feature (deprecated in 5.6 and removed in 6.0 - [Breaking Changes 6.0.0](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking-6.0.0-xes.html))
- Cleanup some ansible code (especially in `when` conditions)


test/integration/helpers/serverspec/xpack_spec.rb
```diff
@@ -7,7 +7,9 @@
     it { should_not contain "path.conf: /etc/elasticsearch/#{vars['es_instance_name']}" }
     it { should contain "path.data: /var/lib/elasticsearch/localhost-#{vars['es_instance_name']}" }
     it { should contain "path.logs: /var/log/elasticsearch/localhost-#{vars['es_instance_name']}" }
-    it { should contain 'xpack.security.enabled: false' }
-    it { should contain 'xpack.watcher.enabled: false' }
+    if vars.key?('es_xpack_features')
```
jmlrt: this conditional doesn't seem to work, the 2 following tests are absents of https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/47/OS=ubuntu-1804,TEST_TYPE=xpack-upgrade,VERSION=7.x,label=linux/console even if test with the same conditional in https://github.com/jmlrt/ansible-elasticsearch/blob/627a29ae093285adc3818214b63edb36955368d6/test/integration/helpers/serverspec/shared_spec.rb#L38-L47 is present.

tasks/main.yml
```diff
@@ -68,7 +68,8 @@
 
 - name: set fact manage_native_realm to true
   set_fact: manage_native_realm=true
-  when: es_start_service and (es_enable_xpack and "security" in es_xpack_features) and ((es_users is defined and es_users.native is defined) or (es_roles is defined and es_roles.native is defined))
+  when: es_start_service and (es_enable_xpack and es_xpack_features is defined and "security" in es_xpack_features) and ((es_users is defined and es_users.native is defined) or (es_roles is defined and es_roles.native is defined))
```
Crazybus: Split this out into a multiline list might make it a bit more readable.

tasks/xpack/elasticsearch-xpack.yml
```diff
@@ -20,4 +20,4 @@
 - name: Set elasticsearch.keystore Permissions
   become: yes
   file: state=file path={{ conf_dir }}/elasticsearch.keystore owner={{ es_user }} group={{ es_group }}
-  when: es_enable_xpack and "security" in es_xpack_features and (es_version is version_compare('6.0.0', '>'))
```
Crazybus: I think we can drop the verison compare is > 6 now too.

tasks/xpack/security/elasticsearch-security.yml
```diff
@@ -8,13 +8,13 @@
   file: path={{ conf_dir }}{{ es_xpack_conf_subdir }} state=directory owner={{ es_user }} group={{ es_group }}
   changed_when: False
   when:
-    - es_enable_xpack and "security" in es_xpack_features
+    - es_enable_xpack and es_xpack_features is defined and "security" in es_xpack_features
```
Crazybus: These when conditionals could be moved into the main playbook. Since there is no point calling the elasticsearch-security.yml file if it isn't enabled.

defaults/main.yml
```diff
@@ -29,7 +29,6 @@ es_max_open_files: 65536
 es_max_threads: "{{ 2048 if ( es_version is version_compare('6.0.0', '<')) else 8192 }}"
 es_max_map_count: 262144
 es_allow_downgrades: false
-es_xpack_features: ["alerting","monitoring","graph","ml","security"]
```
Crazybus: It might be easier to just make this an empty list. It would remove the need to check if it is defined all the time and makes the setting a bit more discoverable.

tasks/xpack/elasticsearch-xpack.yml
```diff
@@ -20,4 +21,4 @@
 - name: Set elasticsearch.keystore Permissions
   become: yes
   file: state=file path={{ conf_dir }}/elasticsearch.keystore owner={{ es_user }} group={{ es_group }}
-  when: es_enable_xpack and es_xpack_features is defined and "security" in es_xpack_features and (es_version is version_compare('6.0.0', '>'))
+  when: es_enable_xpack
```
Crazybus: I think we always want to do this even when using the oss version. But it was apparently already like this so its out of scope. I also think that the package properly creates the keystore with the correct permissions now but would need to confirm

tasks/xpack/security/elasticsearch-xpack-activation.yml
```diff
@@ -27,7 +27,7 @@
     return_content: yes
   register: license_activated
   no_log: True
-  when: es_xpack_features is defined and "'security' in es_xpack_features"
+  when: es_api_basic_auth_username is defined or es_api_basic_auth_password is defined
```
Crazybus: I think this should be an `and` rather than `or` for the edge case where someone has only defined the username or password. However I believe at the start of the playbook we detect that both are defined when security is enabled, so things should fail earlier anyway.

Issue #554, opened 2019-04-15
[Feature] 7.0 support


abraxxa:
When will the playbook be ready for 7.0?

wixaw:
We're waiting impatiently :+1:

abraxxa:
Can we please get an update from Elastic on this issue?

jmlrt:
Hi @abraxxa work is in progress to update ansible-elasticsearch to 7.x.
We should have a PR for it soon.

abraxxa:
@jmlrt thanks for the update!

abraxxa:
Great! Will you also do a release of the role?

jmlrt:
@abraxxa The PR merge automatically closed your issue but yes we still need to do the release.

mponton:
Bump. Is the 7.0 support official and soon to be "officially released"? I'm good to pull the latest version and use it but some of my colleagues would feel more comfortable with a tag and official release :smiley:

Thanks!

hjst:
Is there anything that we non-Elastic-co folks can do to help this release? The README says that to get started we should do:

    ansible-galaxy install elastic.elasticsearch,7.0.1

But that obviously results in an error because 7.0.1 isn't tagged. Cloning master _seems_ to work as expected, but the mismatch between the docs & the tagged release over a week after the merge is giving me pause. Is there an unseen testing process running that needs to complete?

jmlrt:
Sorry about the delay for the release and mismatch between README file and tagged release.

We still have 2 PR to add before adding the release tag:

- fix default xpack features during elastic search install (https://github.com/elastic/ansible-elasticsearch/pull/560) related to https://github.com/elastic/ansible-elasticsearch/issues/468
- remove multi-instances support (possibility to install more than one elasticsearch instance in the same machine using different ports, config and datafiles).

Unfortunately, I'm not available to work on it this week and should continue next week.

Meanwhile using the repository master branch is fine and tested.

abraxxa:
The removal of multi-instance support is quite a large change!
Will it handle the changed directory paths?

xixiwangdong:
Why multi-instances support should be removed?
If so, it will be very terrible for me because I just have some servers with huge memory. I can not deploy only one instance per server which is really wasteful.

jmlrt:
Multi-instances support was implemented when it was the only solution available to optimize usage of big servers with a lot of memory as well as allow cluster testing locally.

[...]

PR #566, opened 2019-05-27
Remove multi instances support

jmlrt:
This PR should remove multi-instance support of ansible-elasticsearch role.
As mentionned in https://github.com/elastic/ansible-elasticsearch/issues/554#issuecomment-493253330, with Elasticsearch 7.0 installing more than one node on the same host is no more supported.

⚠️ **It should be merged after https://github.com/elastic/ansible-elasticsearch/pull/567 which is a prerequisite and after new PR to come which should update README instructions to manage es 7.x with multi-instances for existing use cases.**

Note that my IDE cleaned a lot of end of line whitespaces, so I highly recommend selecting GitHub **Hide whitespace changes** checkbox during review :).


README.md
```diff
@@ -15,6 +15,11 @@ Ansible role for 7.x/6.x Elasticsearch.  Currently this works on Debian and RedH
 
 The latest Elasticsearch versions of 7.x & 6.x are actively tested.  **Only Ansible versions > 2.4.3.0 are supported, as this is currently the only version tested.**
 
+**BREAKING CHANGE**
+
+Starting with ansible-elasticsearch:7.0.0, installing more than one instance of Elasticsearch **on the same host** in no more supported.
```
Crazybus: "in no more supported"

"is no longer supported"

README.md
```diff
@@ -234,33 +231,27 @@ An example of a two server deployment is shown below.  The first server holds th
...
     - role: elastic.elasticsearch
   vars:
-    es_instance_name: "node2"
-    es_api_port: 9201
+    ansible_user: ansible
     es_config:
       discovery.seed_hosts: "elastic02:9300"
-      http.port: 9201
-      transport.port: 9301
+      http.port: 9200
```
Crazybus: These are the defaults right? We can just remove them since before they needed to be overriden to allow multiple instances on the same host.

README.md
```diff
@@ -186,23 +185,23 @@ A more complex example:
...
 If the node is deployed to bind on either a different host or port, these must be changed.**
 
 ### Multi Node Server Installations
 
-The application of the elasticsearch role results in the installation of a node on a host. Specifying the role multiple times for a host therefore results in the installation of multiple nodes for the host. 
+The application of the elasticsearch role results in the installation of a node on a host. Specifying the role multiple times for a host therefore results in the installation of multiple nodes for the host.
 
-An example of a two server deployment is shown below.  The first server holds the master and is thus declared first.  Whilst not mandatory, this is recommended in any multi node cluster configuration.  The second server hosts two data nodes.
+An example of a three server deployment is shown below.  The first server holds the master and is thus declared first.  Whilst not mandatory, this is recommended in any multi node cluster configuration.  The two others servers hosts data nodes.
```
Crazybus: Out of scope for this PR. But I think it would be nice to make the example with 3 master nodes, just so that we don't encourage a setup with only a single master.

README.md
```diff
@@ -234,33 +231,27 @@ An example of a two server deployment is shown below.  The first server holds th
...
       node.data: true
       node.master: false
       bootstrap.memory_lock: false
-      cluster.name: "test-cluster"
-    es_scripts: false
     es_templates: false
     es_version_lock: false
-    es_api_port: 9201
-    ansible_user: ansible
+    es_api_port: 9200
```
Crazybus: This one is also a default that can be removed.

README.md
jmlrt: It seems that I make this typo really often 🤦‍♂

README.md
jmlrt: We should left `http.port` as it's a recommendation we have here: 
https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-parameters.yml#L11-L13 (see also https://github.com/elastic/ansible-elasticsearch/pull/558#discussion_r281584111).

But is it still a valid recommendation?

README.md
Crazybus: Haha. It's still there btw :P.

README.md
```diff
@@ -268,14 +320,15 @@ ansible-playbook -i hosts ./your-playbook.yml
 
 ### Installing X-Pack Features
 
-X-Pack features, such as Security, are supported. This feature is currently experimental.
+X-Pack features, such as Security, are supported.
 
-The parameter `es_xpack_features` by default enables all features i.e. it defaults to ["alerting","monitoring","graph","security","ml"]
+The parameter `es_xpack_features` allows to list xpack features to install (example: `["alerting","monitoring","graph","security","ml"]`).
+When the list is empty, it install all features available with the current licence.
```
Crazybus: This variable doesn't actually install anything. The only time `es_xpack_features` is actually referred to is this line:

https://github.com/elastic/ansible-elasticsearch/blob/0e61dba1ce88ecd37e5178f8c54ae8beaa021dd9/tasks/elasticsearch-parameters.yml#L18-L21

Now that it is empty by default I'm not sure if this adds any value at all. Instead I think we can just drop this whole variable and make sure that we explain that you need to set the username and password if you are using security.

tasks/xpack/security/elasticsearch-xpack-activation.yml
```diff
@@ -9,7 +9,7 @@
     return_content: yes
   register: license_activated
   no_log: True
-  when: not "security" in es_xpack_features
```
Crazybus: This is another task where we could use omit to combine them into a single task. Not blocking for this PR though.

test/integration/helpers/serverspec/shared_spec.rb
```diff
@@ -35,12 +51,34 @@
...
+      if vars.key?('es_xpack_features')
+        curl_json("#{es_api_url}/_xpack", username=username, password=password)['features'].each do |feature,values|
+          enabled = vars['es_xpack_features'].include? feature
+          status = if enabled then 'enabled' else 'disabled' end
+          it "the xpack feature '#{feature}' to be #{status}" do
+            expect(values['enabled'] = enabled)
+          end
+        end
+      else
+        features.each do |feature, status|
```
Crazybus: Whats the idea of wanting to test this? If this were to start failing it would be because of a changed default in an Elasticsearch version. If that were to happen we would just be updating the defaults in this test file right?

test/integration/oss-upgrade.yml
```diff
@@ -6,7 +6,7 @@
   roles:
     - elasticsearch
   vars:
-    es_version: "{{ '7.0.0' if es_major_version == '7.x' else '6.7.1' }}" # This is set to an older version than the current default to force an upgrade
+    es_version: "{{ '7.0.1' if es_major_version == '7.x' else '6.7.2' }}" # This is set to an older version than the current default to force an upgrade
```
Crazybus: This won't be a problem when we bump to 7.1.0. However setting it to 7.0.1 means that its not actually testing the version upgrade because the default in this branch is also 7.0.1

README.md
```diff
@@ -13,22 +13,92 @@ Ansible role for 7.x/6.x Elasticsearch.  Currently this works on Debian and RedH
...
+## Dependency
 This role uses the json_query filter which [requires jmespath](https://github.com/ansible/ansible/issues/24319) on the local machine.
 
 ## Usage
 
 Create your Ansible playbook with your own tasks, and include the role elasticsearch. You will have to have this repository accessible within the context of playbook.
 
 ```sh
-ansible-galaxy install elastic.elasticsearch,7.0.1
+ansible-galaxy install git+https://github.com/elastic/ansible-elasticsearch.git,7f5be969e07173c5697432141e909b6ced5a2e94
```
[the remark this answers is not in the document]
jmlrt: This commit is fine. There is only https://github.com/elastic/ansible-elasticsearch/commit/25bd09f6835b476b6a078676a7d614489a6739c5 after this one with no change in ansible code. Morever, this will be changed in next PR with changelog and git tag release.

test/integration/oss-upgrade.yml
jmlrt: Yes need to change that. My search & replace for checking versions to update was a bit too much.

README.md
jmlrt: created https://github.com/elastic/ansible-elasticsearch/issues/571 to follow up

README.md
jmlrt: created https://github.com/elastic/ansible-elasticsearch/issues/572 to follow up

test/integration/helpers/serverspec/shared_spec.rb
jmlrt: The idea was to still have a test when `es_xpack_features` variable is empty but I can remove it if you think it's not relevant.

jmlrt: strange that you saw a change in this PR because this is a change from https://github.com/elastic/ansible-elasticsearch/pull/560/files which is already merged

README.md
jmlrt: It's strange that you saw a change in this PR because this is a change from https://github.com/elastic/ansible-elasticsearch/pull/560/files which is already merged

PR #568, opened 2019-05-29
skip java install on 7.x (java embeded in 7.x version)

jmlrt:


defaults/main.yml
```diff
@@ -11,7 +11,6 @@ es_apt_key: "{{ es_repo_base }}/GPG-KEY-elasticsearch"
 es_apt_url: "deb {{ es_repo_base }}/packages/{{ es_repo_name }}/apt stable main"
 es_apt_url_old: "deb http://packages.elastic.co/elasticsearch/{{ es_repo_name }}/debian stable main"
 es_start_service: true
-es_java_install: true
```
Crazybus: I think setting the default to `false` and then only overriding it for 6.x would be a lot cleaner and clearer. Saves needing to check if the variable is defined and removes the task for setting it to false for 7.x.

PR #569, opened 2019-05-29
Small fixes and improvements

jmlrt:
- add .ruby-version
- update ansible min version
- update ansible galaxy tags
- fix doc and changelog


README.md
```diff
@@ -23,7 +32,7 @@ This role uses the json_query filter which [requires jmespath](https://github.co
 Create your Ansible playbook with your own tasks, and include the role elasticsearch. You will have to have this repository accessible within the context of playbook.
 
 ```sh
-ansible-galaxy install elastic.elasticsearch,7.0.1
+ansible-galaxy install elastic.elasticsearch,7.0-multi
```
Crazybus: I don't think this version tag will work for ansible galaxy. They ignore anything that isn't semver compliant https://galaxy.ansible.com/docs/contributing/version.html. So it will need to be something like 7.0.1-multi

PR #575, opened 2019-06-05
Docs: Fix name of elasticsearch ansible role

flyinggecko:
Use correct role name from ansible galaxy


jmlrt:
jenkins test this please

Issue #570, opened 2019-05-31
Python3 "'dict object' has no attribute \"dict_keys(['... "


svanschalkwyk:
Env: Python 3.7.1
ansible-elasticsearch
elasticsearch-security-file.yml, line 73:
```"#Set passwords for all users declared - Required as the useradd will not change existing user passwords
- name: Set User Passwords
  become: yes
  command: >
    {{es_home}}/bin/{{es_xpack_users_command}} passwd {{ item }} -p {{es_users.file[item].password}}
  with_items: "{{ es_users.file.keys() | default([]) }}"
  when: manage_file_users
 ---
```
```hosts snippet:
es_users:
      file: 
        steph:
          password: 'xxxx'
          roles:
            - admin
```
Fails with Python3 (`python3 $(which ansible-playbook) -i hosts-prod  ./7-elasticsearch_master.yml`)
`'dict object' has no attribute \"dict_keys(['steph'])`
Succeeds in Python2

jmlrt:
Hi @svanschalkwyk,
We don't currently test this role with Python3 but can take a look.

What version of Ansible are you using?

svanschalkwyk:
ansible 2.9.0.dev0

jmlrt:
This is related to some change in dictionaries in Python3: https://docs.ansible.com/ansible/2.4/playbooks_python_version.html#dictionary-views

PR #578, opened 2019-06-06
fix ''dict object has no attribute dict_keys" issue with python3

jmlrt:
In Python2, the dict.keys(), dict.values(), and dict.items() methods returns a list. Jinja2 returns that to Ansible via a string representation that Ansible can turn back into a list. In Python3, those methods return a dictionary view object.
resource: https://docs.ansible.com/ansible/2.4/playbooks_python_version.html#dictionary-views

Fix https://github.com/elastic/ansible-elasticsearch/issues/570

### DIFF
--- a/tasks/xpack/security/elasticsearch-security-file.yml
+++ b/tasks/xpack/security/elasticsearch-security-file.yml
@@ -1,6 +1,6 @@
 ---
 - name: set fact manage_file_users
-  set_fact: manage_file_users=es_users is defined and es_users.file is defined and es_users.file.keys() | length > 0
+  set_fact: manage_file_users=es_users is defined and es_users.file is defined and es_users.file | list | length > 0
 
 - name: Check if old users file exists
   stat:
@@ -35,7 +35,7 @@
   check_mode: no
 
 - name: set fact users_to_remove
-  set_fact: users_to_remove={{ current_file_users.stdout_lines | difference (es_users.file.keys()) }}
+  set_fact: users_to_remove={{ current_file_users.stdout_lines | difference (es_users.file.keys() | list) }}
   when: manage_file_users
 
 #Remove users
@@ -51,7 +51,7 @@
     ES_HOME: "{{es_home}}"
 
 - name: set fact users_to_add
-  set_fact: users_to_add={{ es_users.file.keys() | difference (current_file_users.stdout_lines) }}
+  set_fact: users_to_add={{ es_users.file.keys() | list | difference (current_file_users.stdout_lines) }}
   when: manage_file_users
 
 #Add users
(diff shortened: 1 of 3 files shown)


tasks/xpack/security/elasticsearch-security-file.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: set fact manage_file_users
-  set_fact: manage_file_users=es_users is defined and es_users.file is defined and es_users.file.keys() | length > 0
```
Crazybus: Why are some of these now `.keys() | list` while others are just `list` with the `.keys()` also removed? I can't figure out why some still need `.keys()` and some don't.

tasks/xpack/security/elasticsearch-security-file.yml
```diff
@@ -72,7 +72,7 @@
   become: yes
   command: >
     {{es_home}}/bin/{{es_xpack_users_command}} passwd {{ item }} -p {{es_users.file[item].password}}
-  with_items: "{{ es_users.file.keys() | default([]) }}"
```
Crazybus: Can the `default([])` be removed here? Does list automatically return an empty list if the variable is undefined? 

It fails for me on https://ansible.sivel.net/test/ but maybe it doesn't handle empty variables properly.

tasks/xpack/security/elasticsearch-security-file.yml
jmlrt: just a mistake => fixed

tasks/xpack/security/elasticsearch-security-file.yml
jmlrt: With the `when: manage_file_users` and `set_fact: manage_file_users=es_users is defined and es_users.file is defined and es_users.file.keys() | list | length > 0` at the top of the file, if the variable is undefined we will skip this task so I think we can avoid having an additional `default` filter here, WDYT.

tasks/xpack/security/elasticsearch-security-file.yml
Crazybus: There were a few more were it was removed. Which is why I thought it was on purpose. At least 2 more in this PR where it was deleted.

tasks/xpack/security/elasticsearch-security-file.yml
Crazybus: Makes sense to me!

tasks/xpack/security/elasticsearch-security-file.yml
jmlrt: Are you sure? I don't see no more `| list` without `.keys()`

svanschalkwyk:
Thanks! Quick turn-around!

PR #588, opened 2019-06-28
Set userid and groupid in a different place in the role

broferek:
Hi,

Setting the userid and the groupid is good in the elasticsearch.yml tasks file when you start a fresh new installation.

But in the case of a migration from 6.8 to 7.2 for example, the role will remove everything related to 6.8 version and therefore remove the elasticsearch user and group that you have previously forced and the installation of the 7.2 will create a new user and a new group not taking into account your IDs. This will mess up all your directories ownerships (data, logs, pid, ...)

This PR creates the userid and the groupid between the uninstallation and the installation of Elastic so that you keep your existing IDs no matter if you are installing a fresh new cluster or upgrading an existing one.

Regards


broferek:
@jmlrt can someone review this? 

The current role breaks everything if you use custom user and group ids and wants to upgrade a running 6.8 cluster to 7.2.

jmlrt:
@broferek,
Can you post a sample of playbooks and procedure you use to install es then upgrade it? 

I don't understand how this can delete user and change files uid/gid and we didn't see this during our tests but I'm really interrested to have more details.

broferek:
The problem comes from the order of the tasks.

In tasks/elasticsearch.yml, the user and group is set at the very beginning 

When you start your first installation, it is fine, as the package will be installed afterwards.

But, when you start your upgrade, it will set the user and group again, which is idempotent and fine but the next included 'tasks/elasticsearch-Redhat.yml' will perform an uninstallation of the previous version and in the RPM, it will remove the user and group you previously created, then it installs the new version which will create the user and group again with another IDs.

My fix consists in creating the user between the uninstallation and the installation, in this case, you always end up with the same IDs.

Install 6.8 playbook :
`- hosts: "{{ hosts }}"
  gather_facts: true
  roles:  
    - role: galaxy_elasticsearch
      es_instance_name: "foo" 
      es_config:
        node.name: "foo"
        cluster.name: "bar"
        discovery.zen.ping.unicast.hosts: "localhost:9300"
        http.port: 9200
        transport.tcp.port: 9300
        node.data: true
        node.master: true
        bootstrap.memory_lock: true
        network.host: "{{ ansible_default_ipv4.address }}"
      vars:
        es_enable_xpack: false
        es_version: "6.8.0"
        es_scripts: false
        es_templates: false
        es_version_lock: false
        es_heap_size: 16g
        es_api_port: 9200
        es_api_host: "{{ inventory_hostname }}"
        es_user_id: "1005"
        es_group_id: "1005"`

Install 7.2 playbook :
`- hosts: "{{ hosts }}"
  gather_facts: True
  roles:  
    - role: galaxy_elasticsearch_7
      es_instance_name: "foo" 
      es_config:
        node.name: "foo"
        cluster.name: "bar"
        discovery.seed_hosts: "localhost:9300"
        http.port: 9200
        transport.port: 9300
        node.data: true
        node.master: true
        bootstrap.memory_lock: true
        network.host: "{{ ansible_default_ipv4.address }}"
      vars:
        es_version: "7.2.0"
        es_heap_size: 16g
        es_api_host: "{{ inventory_hostname }}"
        es_data_dirs:
          - "/var/lib/elasticsearch/{{ inventory_hostname }}-foo"
        es_log_dir: "/var/log/elasticsearch/{{ inventory_hostname }}-foo"
        es_pid_dir: "/var/run/elasticsearch/{{ inventory_hostname }}-foo"
        es_conf_dir: "/etc/elasticsearch/foo"
        es_user_id: "1005"
        es_group_id: "1005"`

jmlrt:
`es_enable_xpack` is set to false in 6.8 playbook but not define (true by default) in 7.2 playbook.
Is it expected to not only upgrade from 6.8 to 7.2 but also switch from oss version to elastic license version?

broferek:
Yes, as I want to benefit from xpack features.

jmlrt:
jenkins test this please

jmlrt:
@broferek OK that makes more sense now, thank you for the details.

broferek:
The build failed but it is not linked to my change and it seems it failed on this task since several builds.

jmlrt:
@broferek the build failure is not related to your pr but I can't merge without this test.
I'll try to find some time to fix the build next week.

Crazybus:
@jmlrt since this is also failing in master I don't see any problem with merging this in. 

I have taken a closer look at the failure in master. The actual issue is that the elastic password is not being updated (which causes the failure when it tries to authenticate with the new password to install the license). 

When I try to do it directly with curl I get:

```
root@71565e081141:~# curl -u elastic:changeme localhost:9200/_xpack/security/user/elastic/_password -XPOST -H 'Content-Type: application/json' -d'{"password": "elasticChanged"}
'
{"error":{"root_cause":[{"type":"status_exception","reason":"Cluster state has not been recovered yet, cannot write to the security index"}],"type":"status_exception","reason":"Cluster state has not been recovered yet, cannot write to the security index"},"status":503}
```

My first bet is that we are trying to upgrade from an older version of 6.x which doesn't work with 7.1

Crazybus:
Looking back at the history of the master build it started failing the moment that I updated the test license. 

```
Caused by: java.lang.IllegalStateException: Transport TLS ([xpack.security.transport.ssl.enabled]) is required for license type [platinum] when security is enabled
```

Is the error in the elasticsearch logs after activating it. I have updated it to the correct license so it should start working again.

Crazybus:
jenkins test this please

Issue #576, opened 2019-06-05
[7.1.1] fails on version 6 node with authentication


abraxxa:
**Elasticsearch version**: 6.7.1

**Role version**:  7.1.1

**JVM version** (`java -version`): openjdk-11-jre-headless:amd64 11.0.3+7-1ubuntu2~18.04.1 from Ubuntu 18.04 LTS

**OS version** (`uname -a` if on a Unix-like system): Linux hostname 4.18.0-18-generic #19~18.04.1-Ubuntu SMP Fri Apr 5 10:22:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

**Description of the problem including expected versus actual behaviour**:
When using the 7.1.1 playbook to deploy a 6.7.1 cluster node with a license the activation fails with no useful error message because of no_log: True.
Commenting it out shows:

> FAILED! => {"msg": "The conditional check 'license_activated.status != 200 or license_activated.json.license_status is not defined or license_activated.json.license_status != 'valid'\n' failed. The error was: error while evaluating conditional (license_activated.status != 200 or license_activated.json.license_status is not defined or license_activated.json.license_status != 'valid'\n): 'dict object' has no attribute 'status'"}

Commenting out the failed_when lines shows:

> FAILED! => {"changed": false, "msg": "AbstractDigestAuthHandler does not support the following scheme: 'Bearer'"}

When re-adding `force_basic_auth: yes`, which was removed in commit 2cb020a4c227710d29d0fb71004ccf203e608dec, it succeeds.

The same happens for 'Install templates' in tasks/elasticsearch-template.yml with the same fix.

Not sure if you're already testing the combination version 6.x with authentication but now that 6.8 has it even in the free version it would make sense, especially for paying upgraders like us.

Thanks!

Crazybus:
Thanks for reporting this! I have opened https://github.com/elastic/ansible-elasticsearch/pull/591 to add it back in. I also checked and found out that all other http requests still have this set.

> Not sure if you're already testing the combination version 6.x with authentication but now that 6.8 has it even in the free version it would make sense, especially for paying upgraders like us.

We actually do, this is what the [current testing matrix](https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+master/) looks like:

<img width="620" alt="image" src="https://user-images.githubusercontent.com/4550136/60917167-97f64400-a290-11e9-9d78-aba4b170644c.png">

[...]

PR #591, opened 2019-07-09
Add back in force_basic_auth for all http requests

Crazybus:
Related: #576
This is needed when using security for 6.x. All other http requests
still have this parameter added.


abraxxa:
I can confirm that this is fixed in the 7.4.0 release of the role.

PR #594, opened 2019-07-21
Remove always tag

Scriptkiddi:
The usage of the always tag leads to errors if you run your playbook with a task that only gathers facts once, because always is executed even if the run is limited to another host.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -6,18 +6,12 @@
 
 - name: os-specific vars
   include_vars: "{{ansible_os_family}}.yml"
-  tags:
-      - always
 
 - name: set compatibility variables
   include: compatibility-variables.yml
-  tags:
-      - always
 
 - name: check-set-parameters
   include: elasticsearch-parameters.yml
-  tags:
-      - always
 
 - name: use snapshot release
   include: snapshot-release.yml


jmlrt:
jenkins test this please

jmlrt:
Hi @Scriptkiddi,

The `always` tags is used here to be sure that these tasks will always be run if we run only part of the role with `--tags java` or `--tags config` for example because the role will fail if these prerequisites task are not run. So I think we should keep the `always` tags in the role.

I your specific case, maybe the best thing would be to run your playbook with `--skip-tags always`.

Scriptkiddi:
Hm I would argue for removing the always tag, because even if the role should be skipped those taks are executed, which in my opinion is unexpected behaviour. I would argue for adding the other tags instead to each task

jmlrt:
Can you post your playbook and give us more details about why you want to run it while skipping elasticsearch role and how you are doing it and why running gather facts on elasticsearch hosts is source of errors?

Don't hesitate to post commands and errors which could help us understand your use case.

Scriptkiddi:
We auto generate our playbooks by gathering all dependencies for each host. You are looking at the resulting playbook. 
[playbook.txt](https://github.com/elastic/ansible-elasticsearch/files/3417339/playbook.txt)

```yaml
- become: true
  gather_facts: false
  hosts:
  - elasticsearch
  name: Execute role elasticsearch
  pre_tasks:
  - name: Gather facts
    setup: {}
    when: not ansible_facts
  roles:
  - elasticsearch
  tags:
   - _elasticsearch
```

The problem now is that if you run this playbook all roles are expanded for each host. so if you have a host that is not running elasticsearch, the tasks marked with always are still executed. This should also lead to problems for other inventories that run a playbook that includes hosts that do not run elasticsearch because the always tasks are executed regardless if the host actually runs elasticsearch

jmlrt:
Do you mean that there are hosts in your `elasticsearch` group that are not running elasticsearch?

Scriptkiddi:
no you can have host not in the elasticsearch group, but the role included in your playbook and the tasks will still run in the elasticsearch role

SlothOfAnarchy:
I work in on same project. This is the problem we encouter when executing only specific tags, but do not restrict the hosts with `-l`: 
```
master $ ansible-playbook --tags=_oracle-java,_kibana playbooks/test.yml

PLAY [Execute role oracle-java] ***********************************************************************************************************************************************************************

TASK [Gather facts] ***********************************************************************************************************************************************************************************
ok: [wiki01]
ok: [ticket01]
ok: [build01]

TASK [oracle-java : Add apt key] **********************************************************************************************************************************************************************
ok: [wiki01]
ok: [ticket01]
ok: [build01]

TASK [oracle-java : Add apt repository] ***************************************************************************************************************************************************************
ok: [wiki01]
ok: [build01]
ok: [ticket01]

TASK [oracle-java : Accept Java license] **************************************************************************************************************************************************************
ok: [wiki01]
ok: [ticket01]
ok: [build01]

TASK [oracle-java : Install Java] *********************************************************************************************************************************************************************
ok: [wiki01]
ok: [ticket01]
ok: [build01]

TASK [oracle-java : Install custom certificates] ******************************************************************************************************************************************************
ok: [wiki01] => (item=/etc/ssl/ldapCA.pem)
ok: [ticket01] => (item=/etc/ssl/ldapCA.pem)
ok: [build01] => (item=/etc/ssl/ldapCA.pem)

PLAY [Execute role elasticsearch] *********************************************************************************************************************************************************************

TASK [elasticsearch : os-specific vars] ***************************************************************************************************************************************************************
fatal: [elasticsearch01]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'ansible_os_family' is undefined

The error appears to have been in '/home/michel/stuvus_config/roles/elasticsearch/tasks/main.yml': line 7, column 3, but may

The offending line appears to be:


- name: os-specific vars
  ^ here


fatal: [elasticsearch02]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'ansible_os_family' is undefined

The error appears to have been in '/home/michel/stuvus_config/roles/elasticsearch/tasks/main.yml': line 7, column 3, but may

The offending line appears to be:


- name: os-specific vars
  ^ here
PLAY RECAP ********************************************************************************************************************************************************************************************
build01                    : ok=6    changed=0    unreachable=0    failed=0
elasticsearch01            : ok=0    changed=0    unreachable=0    failed=1
elasticsearch02            : ok=0    changed=0    unreachable=0    failed=1
ticket01                   : ok=6    changed=0    unreachable=0    failed=0
wiki01                     : ok=6    changed=0    unreachable=0    failed=0
```

playbooks/test.yml:
```
- become: true
  gather_facts: false
  hosts:
  - oracle-java
  name: Execute role oracle-java
  pre_tasks:
  - name: Gather facts
    setup: {}
    when: not ansible_facts
  roles:
  - oracle-java
  tags:
  - _oracle-java
- become: true
  gather_facts: false
  hosts:
  - elasticsearch
  name: Execute role elasticsearch
  pre_tasks:
  - name: Gather facts
    setup: {}
    when: not ansible_facts
  roles:
  - elasticsearch
  tags:
  - _elasticsearch
- become: true
  gather_facts: false
  hosts:
  - kibana
  name: Execute role kibana
  pre_tasks:
  - name: Gather facts
    setup: {}
    when: not ansible_facts
  roles:
  - kibana
  tags:
  - _kibana
```

jmlrt:
@Scriptkiddi @SlothOfAnarchy, could you try your playbook with `gather_facts: true` instead of `gather_facts: false` then `setup: {} when: not ansible_facts`? I think that may be the cause of your issue.

cc @Crazybus for any opinion on this issue

Scriptkiddi:
Even if I switch to this part in the playbook, I get the a similar error running it.
see blow for palybook and logs.
The problem as far as I understand it is that, the elasticsearch role executes the always tasks because of the tags even if the role is not supposed to be executed at all since i specificed a different tag for the playbook 
```
ansible-playbook --tags _rsync_backup all
```
```yaml
- become: true
  gather_facts: true
  hosts:
  - elasticsearch
  name: Execute role elasticsearch
  roles:
  - elasticsearch
  tags:
  - _elasticsearch
```

```
PLAY [Execute role elasticsearch] *************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ************************************************************************************************************************************************************************************************************************
ok: [elasticsearch01]
ok: [elasticsearch02]

TASK [elasticsearch : os-specific vars] *******************************************************************************************************************************************************************************************************
ok: [elasticsearch01]
ok: [elasticsearch02]

TASK [elasticsearch : Set the defaults here otherwise they can't be overriden in the same play if the role is called twice] *******************************************************************************************************************
fatal: [elasticsearch01]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'es_major_version' is undefined

The error appears to have been in '/xxxxx/roles/elasticsearch/tasks/compatibility-variables.yml': line 9, column 3, but may

The offending line appears to be:

```

SlothOfAnarchy:
> @Scriptkiddi @SlothOfAnarchy, could you try your playbook with `gather_facts: true` instead of `gather_facts: false` then `setup: {} when: not ansible_facts`? I think that may be the cause of your issue.

Setting `gather_facts: false` is intended for our setup, because it speeds up the playbook execution immensely when facts are only gathered for the hosts actually executed (this happens with the `pre_task` that checks if facts have been gathered yet).

Crazybus:
In case it wasn't clear, these `always` tags have been added so that users can run ansible in [check mode](https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html). This is a valid and normal use case that we want to support. Removing this tag would break a normal and recommended workflow. It seems like you are using an unconventional workflow that is not compatible with roles supporting check mode. 

> Setting gather_facts: false is intended for our setup, because it speeds up the playbook execution immensely when facts are only gathered for the hosts actually executed (this happens with the pre_task that checks if facts have been gathered yet).

One way to speed up fact gathering is to [enable caching](https://docs.ansible.com/ansible/latest/plugins/cache.html#enabling-fact-cache-plugins).

> `ansible-playbook --tags _rsync_backup all`

Have you got this `_rsync_backup` tag inside the same playbook as the ansible role? These sound like two different plays that should be in separate playbooks.

SlothOfAnarchy:
> In case it wasn't clear, these `always` tags have been added so that users can run ansible in [check mode](https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html). This is a valid and normal use case that we want to support. Removing this tag would break a normal and recommended workflow. It seems like you are using an unconventional workflow that is not compatible with roles supporting check mode.
> 
> > Setting gather_facts: false is intended for our setup, because it speeds up the playbook execution immensely when facts are only gathered for the hosts actually executed (this happens with the pre_task that checks if facts have been gathered yet).
> 
> One way to speed up fact gathering is to [enable caching](https://docs.ansible.com/ansible/latest/plugins/cache.html#enabling-fact-cache-plugins).
> 
> > `ansible-playbook --tags _rsync_backup all`
> 
> Have you got this `_rsync_backup` tag inside the same playbook as the ansible role? These sound like two different plays that should be in separate playbooks.

Yes, all our roles are in the same playbook to ensure that they're being executed in the correct order (some roles depend on others)

Have you looked at my alternative proposal (https://github.com/elastic/ansible-elasticsearch/pull/596) which would only add a condition to the tasks? This would work for us and I think your use case would still work with it.

jmlrt:
@SlothOfAnarchy I'm sorry but https://github.com/elastic/ansible-elasticsearch/pull/596 could be source of issues in the future with the standard use cases of this ansible-role. Your use case is really specific and not necessarily following Ansible best practices and we don't want to "complexify" our ansible code to support this kind of use case.

>Yes, all our roles are in the same playbook to ensure that they're being executed in the correct order (some roles depend on others)

I would suggest using [Role Dependencies](https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-dependencies) and split your plays into dedicated playbooks.

If that's not possible, you can still run the playbook with `--skip-tags always` as already mentioned.

Issue #585, opened 2019-06-26
LimitMEMLOCK=infinity no longer implemented


broferek:
<!

**Elasticsearch version** : 7.1.1

**Role version**:  7.1.1

**JVM version** (`java -version`): openJDK 1.8.0

**OS version** (`uname -a` if on a Unix-like system): Centos 7

**Description of the problem including expected versus actual behaviour**: 
With 6.6 role, the systemd service template is deployed by the role and overload the setting LimitMEMLOCK=infinity. This behaviour disappeared with the 7.1.1 role and the service can't start because it can not lock the memory.

Is there another preconisation to work around this problem? Or shall the template be deployed again by the role?

Regards

broferek:
@jmlrt I can propose a PR with a fix for this with a lineinfile task to add it if memory lock is enabled and the server uses systemD but if you can give me the reason why it has been removed, maybe it can help me finding another workaround

jmlrt:
Hello @broferek,

Thanks for this issue.

With OS using systemd, this should be configured in `/etc/systemd/system/elasticsearch.service.d/override.conf` (see [elasticsearch.j2#L51-L59](https://github.com/elastic/ansible-elasticsearch/blob/master/templates/elasticsearch.j2#L51-L59) and [Systemd configuration](https://www.elastic.co/guide/en/elasticsearch/reference/current/setting-system-settings.html#systemd)).

However, this should be added only in addition of [bootstrap.memory_lock](https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-configuration-memory.html#bootstrap-memory_lock) and only when user can't disable swap (see https://github.com/elastic/elasticsearch/issues/22734).

It would be great if you want to create a PR for that. 

`LimitMEMLOCK=infinity` should be set only when `m_lock_enabled` fact is set in [elasticsearch-parameters.yml#L33-L34](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/elasticsearch-parameters.yml#L33-L34).

broferek:
I'm on vacation but will propose a PR according to your advices for that when I'm back.

jmlrt:
Thanks, I wish you good holidays then 😃

PR #595, opened 2019-07-22
Set limitMEMLOCK for OS using systemd

broferek:
This PR fixes https://github.com/elastic/ansible-elasticsearch/issues/585

@jmlrt  I guess I respected your recommendations.

### DIFF
--- a/tasks/elasticsearch-config.yml
+++ b/tasks/elasticsearch-config.yml
@@ -25,16 +25,17 @@
   notify: restart elasticsearch
 
 #Copy the systemd specific file if systemd is installed
-- name: Make sure destination dir exists
-  file: path={{ sysd_config_file | dirname }} state=directory recurse=yes mode=0755
+- when: use_system_d and m_lock_enabled
+  block:
+    - name: Make sure destination dir exists
+      file: path={{ sysd_config_file | dirname }} state=directory recurse=yes mode=0755
   
-- name: Copy Systemd File for Instance
-  become: yes
-  copy: src=systemd/elasticsearch_override.conf dest={{ sysd_config_file }} mode=0644 force=yes
-  when: use_system_d and m_lock_enabled
-  notify:
-  - reload systemd configuration
-  - restart elasticsearch
+    - name: Copy Systemd File for Instance
+      become: yes
+      copy: src=systemd/elasticsearch_override.conf dest={{ sysd_config_file }} mode=0644 force=yes
+      notify:
+       - reload systemd configuration
+       - restart elasticsearch
 
 #Copy the logging.yml
 - name: Copy log4j2.properties File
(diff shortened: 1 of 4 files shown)


jmlrt:
jenkins test this please

tasks/elasticsearch-config.yml
```diff
@@ -24,6 +24,18 @@
   template: src=elasticsearch.j2 dest={{ default_file }} mode=0644 force=yes
   notify: restart elasticsearch
 
+#Copy the systemd specific file if systemd is installed
+- name: Make sure destination dir exists
+  file: path={{ sysd_config_file | dirname }} state=directory recurse=yes mode=0755
```
jmlrt: this task should also be apply only with `when: use_system_d and m_lock_enabled` condition

tasks/elasticsearch-config.yml
```diff
@@ -24,6 +24,18 @@
...
+  
+- name: Copy Systemd File for Instance
+  become: yes
+  copy: src=systemd/elasticsearch_override.conf dest={{ sysd_config_file }} mode=0644 force=yes
```
jmlrt: Maybe we should add [`blockinfile`](https://docs.ansible.com/ansible/latest/modules/blockinfile_module.html#blockinfile-module) instead of `copy` to not override other configuration that some users could add manually to this file.

tasks/elasticsearch-config.yml
broferek: Fixed with https://github.com/elastic/ansible-elasticsearch/pull/595/commits/5994840c75320e460056dd44b8af021d389adf0f

tasks/elasticsearch-config.yml
broferek: I think I found a better solution using ini_file module

This task below will :

- create the file with only this setting if the file doesn't exist
- make sure the setting is defined in the Service section if the file already exists

`ini_file: path={{ sysd_config_file }} section=Service option=LimitMEMLOCK value=infinity mode=0644` 

What do you think about it?

jmlrt: Awesome, I didn't know `ini_file` module.

broferek: Fixed with 5bac5a4

jmlrt:
jenkins test this please

broferek:
@jmlrt any chance to import the last commits into Galaxy ? Galaxy is running a 2 months old version and we use Galaxy for our production roles.

jmlrt:
Well we currently only import commits in Galaxy when we release the ansible roles at the same time as an elastic release. I think we'll do a new release for ansible roles with elastic 7.3.

broferek:
OK, I'll see how I'll deal with this until a new import is done.

jmlrt:
@broferek FYI we've done a new release for 7.4.0 which include this fix.

broferek:
Wonderful ! Thanks for the information

PR #596, opened 2019-07-22
Only run 'always'-tagged tasks if facts have been gathered

SlothOfAnarchy:
The `always` task leads to errors when facts have not been gathered yet on the hosts the role should be executed on. This happens e.g. when another tag limitation excludes the elasticsearch role completely.
This is an alternative for #594, there are also [more details](https://github.com/elastic/ansible-elasticsearch/pull/594#issuecomment-513868280) posted about the bug:


jmlrt:
@SlothOfAnarchy I'm sorry but https://github.com/elastic/ansible-elasticsearch/pull/596 could be source of issues in the future with the standard use cases of this ansible-role. Your use case is really specific and not necessarily following Ansible best practices and we don't want to "complexify" our ansible code to support this kind of use case.

>Yes, all our roles are in the same playbook to ensure that they're being executed in the correct order (some roles depend on others)

I would suggest using [Role Dependencies](https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-dependencies) and split your plays into dedicated playbooks.

If that's not possible, you can still run the playbook with `--skip-tags always` as already mentioned.

SlothOfAnarchy:
The thing is, we want to apply our infrastructure as easily as possible, so I would rather not append `--skip-tags always` to every `ansible-playbook` command that doesn't involve the elasticsearch role. Hardcoding it into an alias would not work because then I can't deploy our elasticsearch hosts with this command anymore.

What kind of possible issues do you see here? Checking for the definition of `ansible_os_family` is just an invariant. If `ansible_os_family` is not defined the task would fail anyway because it directly uses the variable.
Checking it twice after that will also make no difference, because either it is defined and the tasks are executed or it is not defined and the previous task (that uses it directly) would have failed anyways and stopped the execution of the role.

Please reconsider this PR again, otherwise we would have to use a fork of this repo, which is not as easy to update as using your repo directly.

Issue #577, opened 2019-06-06
[7.1.1] don't overwrite log4j2.properties with 6.x non-JSON version


abraxxa:
**Elasticsearch version**: 7.1.1

**Role version**:  7.1.1

**JVM version** (`java -version`): bundled JDK

**OS version** (`uname -a` if on a Unix-like system): Linux hostname 4.18.0-18-generic #19~18.04.1-Ubuntu SMP Fri Apr 5 10:22:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

**Description of the problem including expected versus actual behaviour**:
The template file `templates/log4j2.properties.j2` shipped with the role is from version 6 and logs in plain text and not JSON.
We would prefer an option to not overwrite the file installed by the package at all, for example by setting `es_config_log4j2: null`.

Crazybus:
>  The template file templates/log4j2.properties.j2 shipped with the role is from version 6 and logs in plain text and not JSON.

I think it might even be from version 5 looking at how old the commit messages are :o 

We can just remove it altogether and instead give the user an option to replace the logging config. I believe this was being manually added via a template back when this role supported running multiple instances at the same time.

jmlrt:
@abraxxa, we just merged #637 which stop overwriting log4j2.properties among other things.

abraxxa:
Thanks for the information, I can confirm that this works as expected.

PR #598, opened 2019-07-25
 Template log4j.properties for ES v7 

fabiogermann:
Introduce changes for the logging property file to be compatible with ES v7

### DIFF
--- a/filter_plugins/custom.py
+++ b/filter_plugins/custom.py
@@ -23,7 +23,7 @@ def array_to_str(values=[],separator=','):
 
 def extract_role_users(users={},exclude_users=[]):
     role_users=[]
-    for user,details in users.iteritems():
+    for user,details in users.items():
         if user not in exclude_users and "roles" in details:
             for role in details["roles"]:
                 role_users.append(role+":"+user)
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @fabiogermann, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/598.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

jmlrt:
Hi @fabiogermann,
Thank you for your PR ❤️ 

While updating log4j template for Elasticsearch 7.X is really useful, I wonder if we still need to maintain a specific log4j template into this Ansible role.

After a quick review of the template I don't see any dynamic values inside except conditionals related to Elasticsearch version.

Maybe we should instead keep the `log4j2.properties` file already provided by Elasticsearch packages which is always up to date. WDYT?

jmlrt:
related to https://github.com/elastic/ansible-elasticsearch/issues/577

jmlrt:
I'm closing this PR as we handled it with #637

Issue #499, opened 2018-09-21
es_version_lock: false doesn't remove existing hold


abraxxa:
**Elasticsearch version**
6.4.1

**Role version**:
tag 6.4.0

**JVM version** (`java -version`):
openjdk version "1.8.0_171"
OpenJDK Runtime Environment (build 1.8.0_171-8u171-b11-1~bpo8+1-b11)
OpenJDK 64-Bit Server VM (build 25.171-b11, mixed mode)

**OS version** (`uname -a` if on a Unix-like system):
Linux nac-prod 3.16.0-6-amd64 #1 SMP Debian 3.16.56-1 (2018-04-28) x86_64 GNU/Linux

**Description of the problem including expected versus actual behaviour**:
including elasticsearch-Debian-version-lock.yml is skipped instead of included with different parameters

jmlrt:
Closed by https://github.com/elastic/ansible-elasticsearch/pull/601
Thanks @LukeRoz

PR #601, opened 2019-08-19
Removing package version hold when `es_version_hold: false`

LukeRoz:
Fix for https://github.com/elastic/ansible-elasticsearch/issues/499

### DIFF
--- a/tasks/elasticsearch-RedHat-version-lock.yml
+++ b/tasks/elasticsearch-RedHat-version-lock.yml
@@ -2,6 +2,35 @@
 - name: RedHat - install yum-version-lock
   become: yes
   yum: name=yum-plugin-versionlock state=present update_cache=yes
+
+- name: RedHat - check if requested elasticsearch version lock exists
+  become: yes
+  shell: yum versionlock list | grep -c {{es_package_name}}-{{es_version}}
+  register: es_requested_version_locked
+  args:
+    warn: false
+  failed_when: False
+  changed_when: False
+
 - name: RedHat - lock elasticsearch version
   become: yes
-  shell: yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}{% if es_version is defined and es_version != "" %}-{{ es_version }}{% endif %}
+  shell: yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}-{{ es_version }}
+  args:
+    warn: false
+  when: es_version_lock and es_requested_version_locked.stdout|int == 0
+
+- name: RedHat - check if any elasticsearch version lock exists
+  become: yes
+  shell: yum versionlock list | grep -c elasticsearch
+  register: es_version_locked
+  args:
+    warn: false
+  failed_when: False
+  changed_when: False
+
+- name: RedHat - unlock elasticsearch version
+  become: yes
+  shell: yum versionlock delete 0:elasticsearch*
+  args:
+    warn: false
+  when: not es_version_lock and es_version_locked.stdout|int > 0
--- a/tasks/elasticsearch-Debian-version-lock.yml
+++ b/tasks/elasticsearch-Debian-version-lock.yml
@@ -1,6 +1,0 @@
----
-- name: Debian - hold elasticsearch version
-  become: yes
-  dpkg_selections:
-    name: "{{ es_package_name }}"
-    selection: "hold"
--- a/tasks/elasticsearch-Debian.yml
+++ b/tasks/elasticsearch-Debian.yml
@@ -17,6 +17,13 @@
     changed_when: False
     check_mode: no
 
+  - name: unhold elasticsearch package when switching to a different package type
+    become: yes
+    dpkg_selections:
+      name: "{{ es_other_package_name }}"
+      selection: "install"
+    when: elasticsearch_package.stdout == 'install ok installed'
+
   - name: stop elasticsearch
     service:
       name: 'elasticsearch'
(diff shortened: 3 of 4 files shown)


elasticcla:
Hi @LukeRoz, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/601.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jmlrt:
jenkins test this please

jmlrt:
@LukeRoz test are failing on CentOS with the following error:

```
07:53:52        TASK [elasticsearch : RedHat - check elasticsearch version lock] ***************
07:53:53         [WARNING]: Consider using the yum module rather than running 'yum'.  If you
07:53:53        need to use command because yum is insufficient you can add 'warn: false' to
07:53:53        this command task or set 'command_warnings=False' in ansible.cfg to get rid of
07:53:53        this message.
07:53:53        
07:53:53        fatal: [localhost]: FAILED! => {"changed": true, "cmd": "yum versionlock list | grep -c elasticsearch", "delta": "0:00:00.698033", "end": "2019-09-04 07:53:53.894887", "msg": "non-zero return code", "rc": 1, "start": "2019-09-04 07:53:53.196854", "stderr": "", "stderr_lines": [], "stdout": "0", "stdout_lines": ["0"]}
```

https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/41/OS=centos-7,TEST_TYPE=oss,VERSION=7.x,label=linux/console

tasks/elasticsearch-RedHat-version-lock.yml
```diff
@@ -2,6 +2,19 @@
...
   yum: name=yum-plugin-versionlock state=present update_cache=yes
+
 - name: RedHat - lock elasticsearch version
   become: yes
   shell: yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}{% if es_version is defined and es_version != "" %}-{{ es_version }}{% endif %}
+  when: es_version_lock
+
+- name: RedHat - check elasticsearch version lock
+  become: yes
+  shell: yum versionlock list | grep -c elasticsearch
```
jmlrt: `grep -c elasticsearch` is failing if it doesn't find elasticsearch package.

LukeRoz: I've forgot to add `failed_when: False` and `changed_when: False` to this task. Should be fine now.

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

tasks/elasticsearch-RedHat-version-lock.yml
```diff
@@ -2,6 +2,26 @@
...
+  register: es_version_locked
+  args:
+    warn: false
+  failed_when: False
+  changed_when: False
+
 - name: RedHat - lock elasticsearch version
   become: yes
   shell: yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}{% if es_version is defined and es_version != "" %}-{{ es_version }}{% endif %}
```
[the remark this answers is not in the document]
LukeRoz: Yes, it's necessary to delete old version locks, because they're version specific (obviously :smiley: ) and any existing lock would prevent us from upgrading ES package or setting new version locks during upgrade.

But thanks to pointing that out I've found different problem that would prevent users from performing upgrades via Ansible with version lock engaged. Fix is on the way.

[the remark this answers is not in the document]
jmlrt: My question was more related to `es_version_locked.stdout|int == 0` part of the conditional.

`es_version_locked.stdout` is the result of `yum versionlock list | grep -c {{es_package_name}}-{{es_version}}` so it should be set to the number of elasticsearch packages currently present in versionlock.

As I understand this, If you run `yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}-{{ es_version }}` only when `es_version_locked.stdout|int == 0`, you run it only when there is existing elasticsearch package in versionlock (`yum versionlock list | grep -c {{es_package_name}}-{{es_version}}`), in this case, I don't understand why you need to run `yum versionlock delete 0:elasticsearch*`.

```
- name: RedHat - check elasticsearch version lock
  become: yes
  shell: yum versionlock list | grep -c {{es_package_name}}-{{es_version}}
  register: es_version_locked
  args:
    warn: false
  failed_when: False
  changed_when: False

- name: RedHat - lock elasticsearch version
  become: yes
  shell: yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}-{{ es_version }}
  args:
    warn: false
  when: es_version_lock and es_version_locked.stdout|int == 0
```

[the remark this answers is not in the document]
LukeRoz: I'm running version lock only when requested package is missing from `yum versionlock list`.

Maybe a little example will help:

ES package: `elasticsearch`
installed ES version: `7.0.1`
requested ES version: `7.1.1`
version lock enabled

`yum versionlock list` would return installed ES version:
```
Loaded plugins: fastestmirror, versionlock
0:elasticsearch-7.0.1-1.*
versionlock list done
```
so `yum versionlock list | grep -c elasticsearch-7.1.1`, because requested version is different from installed version, would return:
```
0
```
With this output we will set new version lock for requested version `7.1.1`, but running only `yum versionlock add elasticsearch-7.1.1` would fail with:
```
Excluding 1 update due to versionlock (use "yum versionlock status" to show it)
versionlock added: 0
```
It's necessary to remove existing version lock before creating new one.
I hope that's clear things up a little.

If that would help I can rearrange those tasks that `RedHat - unlock elasticsearch version` would run before `RedHat - lock elasticsearch version` (this time without `yum versionlock delete 0:elasticsearch*`) when upgrade is performed.

tasks/elasticsearch-RedHat-version-lock.yml
```diff
@@ -2,6 +2,35 @@
...
+  shell: yum versionlock list | grep -c elasticsearch
+  register: es_version_locked
+  args:
+    warn: false
+  failed_when: False
+  changed_when: False
+
+- name: RedHat - unlock elasticsearch version
+  become: yes
+  shell: yum versionlock delete 0:elasticsearch* | cat
```
jmlrt: why do you need a `| cat` here?

jmlrt:
jenkins test this please

tasks/elasticsearch-RedHat-version-lock.yml
LukeRoz: Hm... OK, `| cat` is unnecessary now. Deleting.

Issue #607, opened 2019-09-04
Clean specific code related to version < 6.3


jmlrt:
Elasticsearch < 6.3 is no more supported since 2019-08-06 (see [Elastic Product End of Life Dates](https://www.elastic.co/support/eol)).

We have some specific code related to xpack breaking changes in 6.3. (https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/compatibility-variables.yml#L22-L34 for example).

We can now clean this code.

Issue #430, opened 2018-02-20
Permissions on elasticsearch.keystore prevent Elasticsearch from starting


pemontto:
From 6.2, and perhaps earlier, it appears the `elasticsearch.keystore` file is created even if X-Pack security isn't enabled. This then prevents Elasticsearch from starting up
```
Exception in thread "main" org.elasticsearch.bootstrap.BootstrapException: java.nio.file.AccessDeniedException: /etc/elasticsearch/0/elasticsearch.keystore
```

from tasks/xpack/elasticsearch-xpack.yml
```
#Make sure elasticsearch.keystore has correct Permissions
- name: Set elasticsearch.keystore Permissions
  become: yes
  file: state=file path={{ conf_dir }}/elasticsearch.keystore owner={{ es_user }} group={{ es_group }}
  when: es_enable_xpack and "security" in es_xpack_features and (es_version | version_compare('6.0.0', '>'))
```
I'd suggest removing `"security" in es_xpack_features`.
I can create a PR to relax the conditional, and without testing back on 6.x versions ignore errors. Though based on this and other issues, it seems we need some more automated tests around X-Pack features in enabled/disabled states. I will try to look into this when I get some time.

pemontto:
Confirmed the issues exists on 6.1.3 also.

deppfx:
I am running into the same issue with Chef. Were you ever able to get this fixed? I know they set a GUID on `ES_HOME`. Probably setting GUID on `/etc/elasticsearch` should do the trick?

shorsher:
Is there a current work-around so I can start my elasticsearch manually until this is fixed?

pmb311:
@shorsher 

sudo chown root:elasticsearch <ES_PATH_CONF>/elasticsearch.keystore should work.

jasontedor:
Yes, what is needed here is for `chmod g+s ${ES_PATH_CONF}/elasticsearch` after `chown -R root:elasticsearch ${ES_PATH_CONF}`. The `setgid` ensures that when the keystore is created it has the right ownership (`root:elasticsearch`).

sufiyanghori:
Confirm the issue still exists in `6.3.2`, when I tried to start ES after upgrading from `5.6`

klarose:
[...]

PR #609, opened 2019-09-05
No more 6.3 compatibility + Use default files permissions from Elasticsearch package

jmlrt:
- Cleanup legacy code related to Elasticsearch versions < 6.3 => Fix https://github.com/elastic/ansible-elasticsearch/issues/607

- Also use files permissions from official packages instead of changing them => Fix https://github.com/elastic/ansible-elasticsearch/issues/430


tasks/xpack/security/elasticsearch-security-file.yml
```diff
@@ -1,7 +1,10 @@
 ---
-- name: set fact manage_file_users
-  set_fact: manage_file_users=es_users is defined and es_users.file is defined and es_users.file.keys() | list | length > 0
+- set_fact: manage_file_users=false
 
+- set_fact: manage_file_users=true
+  when: es_users is defined and es_users.file is define and es_users.file.keys() | list | length > 0
```
Crazybus:
```suggestion
  when: es_users is defined and es_users.file is defined and es_users.file.keys() | list | length > 0
```

This fixes the currently failing test:

```
14:34:57        TASK [elasticsearch : set_fact] ************************************************
14:34:57        fatal: [localhost]: FAILED! => {"msg": "The conditional check 'es_users is defined and es_users.file is define and es_users.file.keys() | list | length > 0' failed. The error was: template error while templating string: no test named 'define'. String: {% if es_users is defined and es_users.file is define and es_users.file.keys() | list | length > 0 %} True {% else %} False {% endif %}\n\nThe error appears to have been in '/tmp/kitchen/elasticsearch/tasks/xpack/security/elasticsearch-security-file.yml': line 4, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact: manage_file_users=true\n  ^ here\n"}
14:34:57        
```

jmlrt: Fixed in https://github.com/elastic/ansible-elasticsearch/pull/609/commits/08512fc17ea49568a58f788d1ed998133a68fd3e

jmlrt:
@Crazybus regarding the new failing test:

```
         1) Xpack upgrade Tests File "/etc/elasticsearch/users" should be owned by "elasticsearch"
            Failure/Error: it { should be_owned_by 'elasticsearch' }
              expected `File "/etc/elasticsearch/users".owned_by?("elasticsearch")` to return true, got false
              /bin/sh -c stat\ -c\ \%U\ /etc/elasticsearch/users\ \|\ grep\ --\ \\\^elasticsearch\\\$
````

`/etc/elasticsearch/users` file comes from the default RPM with `root` owner.
Should we force `chown` to `elasticsearch` user in this role or let it owned by `root` and remove test on the ownership?

IMO I prefer the 2nd solution. We should keep default RPM settings when we can. WDYT?

Crazybus:
> IMO I prefer the 2nd solution. We should keep default RPM settings when we can. WDYT?

Yup I agree! I think its probably a leftover from the multi instance support.

jmlrt:
@Crazybus This is ready for new review.
I took advantage of it to change files permissions to the ones set by official Elasticsearch packages.

This should also fix https://github.com/elastic/ansible-elasticsearch/issues/430.

PR #617, opened 2019-09-30
use systemd ansible module for daemon-reload

### LINKED ISSUE #498: QA: why is the `reload systemd configuration` needed explicitly
Hey guys, in this file/line:
https://github.com/elastic/ansible-elasticsearch/blob/fb760f4b9008ae046a731ebc0e7d9f3339c14396/handlers/main.yml#L2

reload of the underlying service agent should be handled by the `service` module, so why we need that ?

jmlrt:
does what it says on the tin

fix #498

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,7 +1,8 @@
 
 - name: reload systemd configuration
   become: yes
-  command: systemctl daemon-reload
+  systemd:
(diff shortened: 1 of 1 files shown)


jmlrt:
jenkins test this please

jlozadad:
+1

PR #621, opened 2019-10-14
Fix deprecated URI in x-pack activation task.

diadzine:
URI _xpack/license is deprecated and _license should be use instead.

### DIFF
--- a/tasks/xpack/security/elasticsearch-xpack-activation.yml
+++ b/tasks/xpack/security/elasticsearch-xpack-activation.yml
@@ -2,7 +2,7 @@
 - name: Activate ES license (with security authentication)
   uri:
     method: PUT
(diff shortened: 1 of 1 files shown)


elasticcla:
Hi @diadzine, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/621.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jmlrt:
jenkins test this please

jmlrt:
Hi @diadzine,
Thanks for your PR.
This has been implemented by @pemontto in https://github.com/elastic/ansible-elasticsearch/pull/620/commits/70143c6327d6f84f3ee62032deb77580ee3ca0bf so I'm closing this one.

PR #631, opened 2019-11-14
(#628) Added io.netty.allocator.numDirectArenas=0 flag to jvm.options

andrx:
(#628) Added io.netty.allocator.numDirectArenas=0 flag to overcome RAM issues on ES >= 7.4.0

Issue #628, opened 2019-11-03
Update jvm.options for ES7


andrx:
**Elasticsearch version**
7.4+

**Role version**: 
7.1

**JVM version**
bundled with ES7.4

**OS version** (`uname -a` if on a Unix-like system):
centos7 Linux 3.10.0-957.1.3.el7.x86_64 

**Description of the problem including expected versus actual behaviour**:
I had a question on RAM issues described on [discuss](https://discuss.elastic.co/t/es-6-8-vs-7-4-memory-issues/202849) that led me to check the diff of `jvm.options` between [ansible](https://github.com/elastic/ansible-elasticsearch/blob/master/templates/jvm.options.j2) and [current default shipped with ES7](https://github.com/elastic/elasticsearch/blob/7.4/distribution/src/config/jvm.options)
There are updates that should be synchronized.

**Playbook**:

	- name: Elasticsearch init nodes
	  hosts: es
	  roles:
		- role: elastic.elasticsearch
	  gather_facts: true
	  vars:
		es_major_version: "7.x"
		es_version: "7.4.2"
		es_instance_name: "{{ ansible_env.NODETYPE }}-{{ ansible_env.INSTANCE_INDEX }}"
		es_data_dirs:
		  - "/opt/elasticsearch/data"
		es_log_dir: "/opt/elasticsearch/logs"
		es_config:
		  node.name: "{{ ansible_env.NODETYPE }}-{{ ansible_env.INSTANCE_INDEX }}"
		  cluster.name: "{{ansible_env.CLUSTER_NAME}}"
		  node.attr.role: "{{ansible_env.NODETYPE}}"
		  discovery.seed_hosts: "{{ hostvars[inventory_hostname].groups.es }}"
		  cluster.initial_master_nodes: "{{ hostvars[inventory_hostname].groups.master }}"
		  network.host: "{{ ansible_default_ipv4.address }}"
		  http.port: 9200
		  transport.port: 9300
		  node.data: "{{true if (ansible_env.NODETYPE=='data' or ansible_env.NODETYPE=='shadow') else false}}"
		  node.master: "{{true if ansible_env.NODETYPE=='master' else false}}"
		  node.ml: false
		  bootstrap.memory_lock: true
		  thread_pool.write.queue_size: 1000
		  thread_pool.search.queue_size: 10000
		  indices.queries.cache.size: 5%
		  indices.breaker.total.use_real_memory: false
[...]

PR #400, opened 2017-11-27
Configurable jvm.options.j2 template path

andreausu:

PR #637, opened 2019-11-26
Use default config files

jmlrt:
Lot of ES config files templates provided by this role weren't up to date with files provided by standard elasticsearch 6.x and 7.x packages.

This PR fix that by using standard config file provided by elasticsearch package when possible (`log4j2.properties`) or updating templates to reflect new versions provided by elasticsearch packages (`/etc/sysconfig/elasticsearch`, `/etc/default/elasticsearch` and `jvm.options`).

This is also removing some options deprecated since [6.0](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking_60_packaging_changes.html).

Also allow to fully override these files by providing custom files.

Related to #631, #628, #598, #577 and #400 and https://discuss.elastic.co/t/es-6-8-vs-7-4-memory-issues/202849


jmlrt:
jenkins test this please

jmlrt:
Ubuntu 14.04 failures aren't related => created test-kitchen/kitchen-docker#353 for that

jmlrt:
I have a really strange permissions issue with this PR:

- permissions on master branch:
```
kitchen@3d4e7e477739:~$ ls -ld /etc/elasticsearch/
drwxr-s--- 2 root elasticsearch 4096 Nov 27 16:41 /etc/elasticsearch/
kitchen@3d4e7e477739:~$ ls -ld /var/lib/elasticsearch/
drwxr-s--- 3 elasticsearch elasticsearch 4096 Nov 27 16:41 /var/lib/elasticsearch/
kitchen@3d4e7e477739:~$ ls -ld /var/log/elasticsearch/
drwxr-s--- 2 elasticsearch elasticsearch 4096 Nov 27 16:41 /var/log/elasticsearch/
```

- permissions on this PR:
```
[kitchen@093070b94dce ~]$ ls -ld /etc/elasticsearch/
d-wSrwxrwT 2 root elasticsearch 4096 Nov 27 16:05 /etc/elasticsearch/
[kitchen@093070b94dce ~]$ ls -ld /var/lib/elasticsearch/
d-wSrwxrwT 2 elasticsearch elasticsearch 4096 Oct 28 20:56 /var/lib/elasticsearch/
[kitchen@093070b94dce ~]$ ls -ld /var/log/elasticsearch/
d-wSrwxrwT 2 elasticsearch elasticsearch 4096 Oct 28 20:56 /var/log/elasticsearch/
```

I don't see any change which could give these strange permissions...

Conky5:
@jmlrt maybe try putting all the `mode` values in quotes, the `2750` might be causing this funkiness  https://docs.ansible.com/ansible/latest/modules/file_module.html#file-module
> You must either add a leading zero so that Ansible's YAML parser knows it is an octal number (like 0644 or 01777) or quote it (like '644' or '1777') so Ansible receives a string and can do its own conversion from string into number.

jmlrt:
Finally tests are OK (except ubuntu-1404 due to https://github.com/elastic/ansible-elasticsearch/pull/637#issuecomment-559071500)! @Conky5 @fatmcgav @Crazybus This is ready for review.

tasks/elasticsearch-config.yml
```diff
@@ -4,54 +4,101 @@
...
        - reload systemd configuration
        - restart elasticsearch
 
-#Copy the logging.yml
-- name: Copy log4j2.properties File
+- name: Copy jvm.options File
   become: yes
-  template: src={{ es_config_log4j2 }} dest={{ es_conf_dir }}/log4j2.properties owner=root group={{ es_group }} mode=0660 force=yes
+  template:
+    src: "{% if es_config_jvm != '' %}{{ es_config_jvm }}{% else %}jvm.options.j2{% endif %}"
```
Crazybus: Does this do something different than just setting a default value? So having: `es_config_jvm: "jvm.options.j2"` in `defaults/main.yml`.

tasks/elasticsearch-config.yml
```diff
@@ -4,54 +4,101 @@
...
+    force: yes
   register: system_change
   notify: restart elasticsearch
 
 #Copy the default file
 - name: Copy Default File
   become: yes
-  template: src=elasticsearch.j2 dest={{ default_file }} owner=root group={{ es_group }} mode=0660 force=yes
+  template:
+    src: "{% if es_config_default != '' %}{{ es_config_default }}{% else %}elasticsearch.j2{% endif %}"
```
Crazybus: Same question about whether this should also just be looking at the default value.

templates/jvm.options.j2
```diff
@@ -41,15 +41,37 @@
 -XX:CMSInitiatingOccupancyFraction=75
 -XX:+UseCMSInitiatingOccupancyOnly
 
+## G1GC Configuration
+# NOTE: G1GC is only supported on JDK version 10 or later.
+# To use G1GC uncomment the lines below.
+# 10-:-XX:-UseConcMarkSweepGC
+# 10-:-XX:-UseCMSInitiatingOccupancyOnly
+# 10-:-XX:+UseG1GC
+{% if es_version is version('7.3.0', '<=') %}
```
Crazybus: Is this configuration related to the Elasticsearch version or Java version? I suspect it is related to the Java version which could cause issues for a user that has `es_java_install` set to something other than the default.

Making assumptions about the version is probably ok as long as there is also a way for the user to opt out or override the setting.

templates/log4j2.properties.j2
```diff
@@ -1,85 +0,0 @@
-status = error
```
Crazybus: Is the log4j configuration in the latest 6.x and 7.x releases still the same? Or will this result in changes for some users? 

This is a bit of a tricky change when it comes to backwards compatibility. Deploying the new ansible playbook version to an existing cluster will result in no change since this file becomes un-managed. However if a node is replaced, or a new one is added that single node will end up with a different logging configuration from the rest of the cluster.

templates/jvm.options.j2
Crazybus: I just realized that these lines are all commented out. So maybe the templating should just be removed altogether?

jmlrt: >Is this configuration related to the Elasticsearch version or Java version? 
This configuration is related to Elasticsearch version, it's based on the default [jvm.options] file provided by Elasticsearch RPM/DEB package. I deployed all minor versions from 6.7 to 7.5 on centos/debian containers and compared config files like `jvm.options` to checked all differences when there are some.

Note, that lines prefixed by a number `10-:`, `8:`, ... are relative to only a specific Java version (java 10, java 8, ...) (more details [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/jvm-options.html)).

>I just realized that these lines are all commented out. So maybe the templating should just be removed altogether?

This commented lines are from the default files from RPM/DEB packages, I added them to be the closest has possible to the package files. However we can remove commented lines if you think it's not necessary.

templates/log4j2.properties.j2
jmlrt: The log4j has a lot of changes in 6.x then 7.x versions compared to what was existing in the template we provided.

Indeed, if this file become unmanaged, existing users will keep their old versions and could have inconsistencies in log config, if they deploy new nodes. We could keep managing this file like the other to force it's update for upgrade scenarios.

However, managing these configuration files instead of reusing the ones provided by default packages increase the risk of having deprecated versions as the default files seem to change quiet regularly.

tasks/elasticsearch-config.yml
jmlrt: no differences, fixed in https://github.com/elastic/ansible-elasticsearch/pull/637/commits/ea4d6cf43a8114be9e061ddc5e10afa1c1607a72

tasks/elasticsearch-config.yml
jmlrt: fixed in https://github.com/elastic/ansible-elasticsearch/pull/637/commits/ea4d6cf43a8114be9e061ddc5e10afa1c1607a72

PR #645, opened 2019-12-11
Trigger ca-certificates-java post installation script only when it's installed

ilpianista:
This role fails to execute on a Ubuntu machine with Amazon JDK Corretto because it's assuming `ca-certificates-java` is installed while this isn't the case. 

Amazon JDK Corretto is based on OpenJDK, thus it should be compatible with OpenJDK.

Installing `ca-certificates-java` isn't a workaround as it would install openjdk also (since amazon corretto doesn't provide `java8-runtime-headless`).

### DIFF
--- a/tasks/java.yml
+++ b/tasks/java.yml
@@ -39,16 +39,13 @@
   apt: name={{ java }} state={{java_state}}
   when: ansible_os_family == 'Debian'
 
-- name: register open_jdk version
-  shell: java -version 2>&1 | grep OpenJDK
-  register: open_jdk
-  ignore_errors: yes
-  changed_when: false
-  check_mode: no
+- name: Ubuntu - Get installed packages
+  package_facts:
+  when: ansible_distribution == 'Ubuntu'
 
 #https://github.com/docker-library/openjdk/issues/19 - ensures tests pass due to java 8 broken certs
 - name: refresh the java ca-certificates
   become: yes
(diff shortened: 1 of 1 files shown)


jmlrt:
Hi @ilpianista,
This role manages java JDK only if the [`es_java_install`](https://github.com/elastic/ansible-elasticsearch/blob/master/tasks/main.yml#L30) is set to `true` ([`false`](https://github.com/elastic/ansible-elasticsearch/blob/master/defaults/main.yml#L15) by default with Elasticsearch 7.X as we now [bundle JDK in Elasticsearch distribution](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/release-highlights-7.0.0.html#_bundle_jdk_in_elasticsearch_distribution)).
If you want to use Corretto JDK, you should set `es_java_install` to false.

ilpianista:
> If you want to use Corretto JDK, you should set `es_java_install` to false.

Thanks, I'll do.

Anyway, doesn't still make sense to execute the post installation script only when the package is installed or only when the post installation script itself exists?

jmlrt:
> Anyway, doesn't still make sense to execute the post installation script only when the package is installed or only when the post installation script itself exists?

Well since this task is only run when you let Ansible role manage java install by setting `es_java_install=true` and since `ca-certificates-java` is a dependency of the java JDK package installed by Ansible role, I don't see a use case where this task would be executed while package `ca-certificates-java` wouldn't be installed.

PR #653, opened 2019-12-23
Xpack features refactoring

### LINKED ISSUE #608: Our es_xpack_features variable is broken as it doesn't allow us to select X-Pack features to install anymore
OK after digging for a few hours into Elasticsearch documentation and Ansible code:

Since Elasticsearch 6.3, by default, when you install Elasticsearch, X-Pack is installed (except if you install oss package) (source: https://www.elastic.co/guide/en/elasticsearch/reference/7.1/setup-xpack.html).

There is no way to uninstall X-Pack (source: https://discuss.elastic.co/t/how-can-i-disable-or-uninstall-x-pack-in-elk-stack-version-6-7/175339/2)

If you are not using oss package, the only control that you have on X-Pack features is by enabling or disabling certain specific features in Elasticsearch configuration with `xpack.<FEATURE>.enabled: false`

This setting is available for `monitoring` or `security` for example.

Our `es_xpack_features` variable is broken as it doesn't allow us to select X-Pack features to install anymore and we'll need to address that in another ticket.

Regarding your use case, if you already have Elasticsearch > 6.2 with non oss package, you should already have X-Pack enabled and the upgrade shouldn't have any impact as long as you keep `es_enable_xpack=true`.

Regarding `readonlyrest` plugin interfering with X-Pack, I suspect that disabling security feature with `xpack.security.enabled: false` should be enough.

If you currently have Elasticsearch <= 6.2, then 6.3 was already a breaking change with the integration of X-Pack inside Elasticsearch non oss package and I'm afraid you won't be able to upgrade without any downtime with this ansible role.

_Originally posted by @jmlrt in https://github.com/elastic/ansible-elasticsearch/issues/602#issuecomment-528122756_

jmlrt:
Fixes #608


tasks/elasticsearch-parameters.yml
```diff
@@ -12,10 +12,7 @@
 #Check if working with security we have an es_api_basic_auth_username and es_api_basic_auth_username - otherwise any http calls wont work
 - name: fail when api credentials are not declared when using security
   fail: msg="Enabling security requires an es_api_basic_auth_username and es_api_basic_auth_password to be provided to allow cluster operations"
-  when:
-    - es_enable_xpack and "security" in es_xpack_features
-    - es_api_basic_auth_username is not defined
-    - es_api_basic_auth_password is not defined
+  when: not oss_version and es_api_basic_auth_username is defined and es_api_basic_auth_password is not defined
```
Conky5: Should
```
es_api_basic_auth_username is defined
```
be
```
es_api_basic_auth_username is not defined
```
?

Also I think having the conditions on separate lines is more readable if it is still possible to have that.

defaults/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 es_version: "7.5.1"
 es_use_snapshot_release: false
-es_enable_xpack: true
+oss_version: false
```
Conky5: Could/should we add some kind of backwards compatibility so that if `es_enable_xpack` is set it sets `oss_version` automatically and logs a deprecation message? Perhaps also a deprecation message (or straight up failure) if  `es_xpack_features` is set with instructions on how to achieve this instead (which I think would be settings in `es_config`)?

jmlrt:
still valid

PR #657, opened 2020-01-10
switched relative URLs to absolute URLs 

### LINKED ISSUE #656: Repo internal links in ansible galaxy are broken
The internal Links of the ReadMe in ansible galaxy are broken and result in a 404.

I.e. on top of the ReadMe the `... follow upgrade procedure` points to https://galaxy.ansible.com/docs/multi-instance.md#upgrade-procedure and further down in the `X-Pack Security SSL/TLS` section the link for `... follow the SSL/TLS setup procedure` points to https://galaxy.ansible.com/docs/ssl-tls-setup.md. 
I tried adding the `/elastic/elasticsearch/` to the urls, but that also does not work.
Not sure how ansible galaxy generates those links and if this is something we can fix. 

In the Readme on github both links work fine.

jakommo:
... to fix broken links in Ansible Galaxy.

Fixes #656

@jmlrt I linked to master, which seems like the most easy to maintain, but in theory it could lead to outdated links again in the future if the version on Ansible Galaxy is different than master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ The latest Elasticsearch versions of 7.x & 6.x are actively tested.
 
 ### Notice about multi-instance support
 
-* If you use only one instance but want to upgrade from an older ansible-elasticsearch version, follow [upgrade procedure](./docs/multi-instance.md#upgrade-procedure)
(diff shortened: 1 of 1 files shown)


jmlrt:
Thanks @jakommo, I guess linking to master is OK since the docs aren't moving so much.

Issue #592, opened 2019-07-11
Support the new secure_password for xpack notifications 


MikeBairdRocks:
**Describe the feature**:
when es_mail_config.pass is defined should know the elastic version and generate the secure_password in the keystore if >=7.0.

Docs:
- [Email Notification Settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/notification-settings.html#email-notification-settings)
- [Security Settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/secure-settings.html)

Also leave out the pass in [elasticsearch.yml.j2](https://github.com/elastic/ansible-elasticsearch/blob/master/templates/elasticsearch.yml.j2) since this will prevent 7 from starting with an invalid attribute setting.

Some sudo code:
```yaml
- name: Create Smtp password for notifications
  become: yes
  shell: echo "{{es_mail_config['pass']}}" | {{es_home}}/bin/elasticsearch-keystore add -x 'xpack.notification.email.account.{{ es_mail_config['account'] }}.smtp.secure_password'
  when:
    - es_mail_config['pass'] is defined and list_keystore is defined
  environment:
    ES_PATH_CONF: "{{ es_conf_dir }}"
  no_log: true
```

Crazybus:
Makes a lot of sense. Since there are other secure settings besides this I think it would be a lot cleaner to instead implement a way to populate the keystore in a generic way. So instead of hard-coding it for this specific value just having a way to add files and strings to the keystore in a generic way like:

```
es_keystore:
  files:
    keystore_key: /path/to/file
  strings:
    xpack.notification.email.account.{{ es_mail_config['account'] }}.smtp.secure_password: {{ password_from_somwhere_secure }}
```
```

jmlrt:
Related to https://github.com/elastic/ansible-elasticsearch/pull/503

rs-garrick:
I ran into this problem today. MikeBairdRocks's suggestion looked pretty good.

https://github.com/elastic/ansible-elasticsearch/pull/660

PR #660, opened 2020-01-14
Support the new secure_password for xpack notifications #592

rs-garrick:
I took MikeBairdRocks' idea and made it work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,7 +446,8 @@ To configure X-pack to send mail, the following configuration can be added to th
         host: <mail domain>
         port: <port number>
         user: <e-mail address> --optional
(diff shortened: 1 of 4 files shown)


elasticcla:
Hi @rs-garrick, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ansible-elasticsearch/pull/660.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rs-garrick:
I have amended my commit with the correct email address.

PR #668, opened 2020-01-28
heap.dump.path is set as mandatory variable. Setting it to optional i…

ozonsozons:
heap.dump.path is set as mandatory variable. 
Currently, if heap.dump.path is not defined then after deployment in linux it appears as "-XX:HeapDumpPath=${heap.dump.path}" in elastic+ process

Setting it to optional if defined.

### DIFF
--- a/templates/jvm.options.j2
+++ b/templates/jvm.options.j2
@@ -110,7 +110,9 @@
 
 # specify an alternative path for heap dumps; ensure the directory exists and
 # has sufficient space
--XX:HeapDumpPath=${heap.dump.path}
+{% if heap.dump.path is defined %} 
+-XX:HeapDumpPath=${heap.dump.path} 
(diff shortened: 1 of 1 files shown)


templates/jvm.options.j2
```diff
@@ -110,7 +110,9 @@
 
 # specify an alternative path for heap dumps; ensure the directory exists and
 # has sufficient space
--XX:HeapDumpPath=${heap.dump.path}
+{% if heap.dump.path is defined %} 
+-XX:HeapDumpPath=${heap.dump.path} 
+{% endif %}
```
jmlrt:
```suggestion
${heap.dump.path} 
```

jmlrt:
We will finally add a new Ansible variable `es_heap_dump_path` in #691.
Closing this PR.

PR #669, opened 2020-01-28
Several tasks in elasticsearch-ssl.yml missing become

rs-garrick:
The SSL tasks are all privileged operations. They need `become`.


jmlrt:
jenkins test this please

PR #672, opened 2020-02-10
fix check mode

pgassmann:
fixes error when running `ansible-playbook` with `--check`

Error message was:

`The conditional check 'es_version_locked.stdout|int > 0' failed. The error was: error while evaluating conditional (es_version_locked.stdout|int > 0): 'dict object' has no attribute 'stdout'`


pgassmann:
CLA is signed

jmlrt:
jenkins test this please

PR #678, opened 2020-03-06
Update security task files

nduytg:
Refactor security task files in favor of multi-line format


jmlrt:
jenkins test this please

PR #681, opened 2020-03-19
Introduce es_java_home variable to allow setting JAVA_HOME 

timdev:
This PR introduces a variable called `es_java_home`, which defaults to an empty string.  If set to a non-empty string, that string is used to set the JAVA_HOME environment variable in the default config file (via the `templates/elasticsearch.j2` template).

If `es_java_home` is not changed from its default, the file generated from the template is unchanged.  That is to say, this change should be entirely backwards-compatible.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -47,6 +47,7 @@ es_api_sleep: 15
 es_debian_startup_timeout: 10
 
 # JVM custom parameters
+es_java_home: ''
 es_jvm_custom_parameters: ''
 
 # SSL/TLS parameters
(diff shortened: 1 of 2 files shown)


jmlrt:
Hi @timdev,
Thanks for your PR.
Can you sign the CLA: https://www.elastic.co/contributor-agreement

timdev:
hi @jmlrt - I did so right after submitting the PR.  Let me know if you need me to do so again.

jmlrt:
jenkins test this please

PR #703, opened 2020-06-24
add AMZN Linux 2 support

anisf:
In README.md : "Currently this works on Debian and RedHat based linux systems"

In Elasticsearch documentation : "Prefer the Amazon Linux AMIs"

Amazon-Linux is RedHat based. Starting Amazon-Linux v2, it uses systemd.
But use_system_d var is not set to True (tasks/elasticsearch-parameters.yml: 44) when running this role against Amazon-Linux v2 EC2 instance.

Causing systemd memlock configuration to be skipped because use_system_d needs to be True (tasks/elasticsearch-config.yml: 63)

Causing bootstrap checks failed when es_config['bootstrap.memory_lock'] == True

This PR adds Amazon-Linux v2 to the conditions list to set use_system_d to True


anisf:
WIP: adding tests for amznlinux2

anisf:
tests added

jmlrt:
jenkins test this please

anisf:
@jmlrt 

The problem occurs when you enable bootstrap.memlock and try to deploy on a Amazon Linux v2 instance. Elasticsearch service won't start.

These tests assert the systemd configuration is made when bootstrap.memlock is enabled.

So the Elasticsearch service behaviour is normal, as the bootstrap checks fail as it should. But the role behaviour isn't.

So these tests don't check elasticsearch configuration, but the role itself, which won't work with a specific configuration (and memlock is a best practice) 

What do you think ?

jmlrt:
I understand, however now that systemd config is implemented for amazonlinux in https://github.com/elastic/ansible-elasticsearch/pull/703/files#diff-916074b85953579021fe6cee52c54d8bL45, I don't see a big risk of regression on this part of code.

Once code has been manually tested and merged, the tests are mostly used to ensure default config and some important scenarios involving complex parts of the code base (TLS, authentication, upgrade from previous version or migrate from oss to xpack, ...) continue to work on all platforms and catch regressions before merging new PR.

It would be too costly to create tests for everything and would certainly require that we refactor our CI tests infrastructure to be able to run them all.

jmlrt:
Just thinking about it but can you also update the list of supported OS in the README (https://github.com/elastic/ansible-elasticsearch/blame/master/README.md#L9-L15)?

anisf:
@jmlrt , ok i understand. Tests are now removed and README updated

jmlrt:
jenkins test this please

jmlrt:
That's good for merging. Thank you for improving this role. Thanks to you I also added support for amazonlinux2 in https://github.com/elastic/ansible-beats/pull/89 👍

cjolif:
Thanks a lot for this, any idea on when a release will be issued with that fix/support?

jmlrt:
We don't communicate releases dates, but this is for pretty soon.

Issue #710, opened 2020-08-04
Deleting native user causes all but one node to fail


lksnyder0:
**Elasticsearch version**: 7.8.1

**Role version**:  7.8.1

**JVM version** (`java -version`):
```
# /usr/share/elasticsearch/jdk/bin/java -version
openjdk version "14.0.1" 2020-04-14
OpenJDK Runtime Environment AdoptOpenJDK (build 14.0.1+7)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 14.0.1+7, mixed mode, sharing)
```

**OS version** (`uname -a` if on a Unix-like system): Ubuntu 18.04.3

**Description of the problem including expected versus actual behaviour**:
When multiple nodes are configured and a user is deleted, all nodes but one will fail. I expected any CRUD operations on users and roles to only occur once.

**Playbook**:
```yaml
---
- name: Deploy elasticsearch with testuser
  hosts: elastic_servers
  roles:
    - role: elastic.elasticsearch
  vars:
    es_version: 7.8.1
    es_xpack_trial: true
    initial_master_nodes:
      - node1
      - node2
    es_config:
      node.name: "{{ node_name | default(ansible_hostname) }}"
      cluster.name: "test123"
      discovery.seed_hosts: "xxx.xxx.xxx.xxx:9300, yyy.yyy.yyy.yyy:9300"
      cluster.initial_master_nodes:
        - "node1"
        - "node2"
      network.host: "0.0.0.0"
    es_heap_size: 1g
    es_api_basic_auth_username: elastic
    es_api_basic_auth_password: badpassword
    es_users:
      native:
        testuser:
          password: password123
          roles:
            - superuser
        testuser2:
          password: anotherpassword
          roles:
            - superuser

- name: Deploy elasticsearch without testuser
  hosts: elastic_servers
  roles:
    - role: elastic.elasticsearch
  vars:
    es_version: 7.8.1
    es_xpack_trial: true
    initial_master_nodes:
      - node1
      - node2
    es_config:
      node.name: "{{ node_name | default(ansible_hostname) }}"
      cluster.name: "test123"
      discovery.seed_hosts: "xxx.xxx.xxx.xxx:9300, yyy.yyy.yyy.yyy:9300"
      cluster.initial_master_nodes:
        - "node1"
        - "node2"
      network.host: "0.0.0.0"
    es_heap_size: 1g
    es_api_basic_auth_username: elastic
[...]

PR #716, opened 2020-09-03
Use run_once for api related tasks

lksnyder0:
I am proposing a fix for issue #710. By specifying run_once: True for all tasks related to the native user, roles, and license API. The role can avoid unnecessary API calls as well as fixing the issue. I used the same playbook as in the issue report. The ansible log [here](http://sprunge.us/tJ6xZY) shows a successful run.

This could be further 'optimized' by delegating all the API calls to localhost. However, this would require the es_api_url to be accessible from the ansible machine which may not always be the case.


jmlrt:
jenkins test this please

jmlrt (approved):
LGTM⛴

PR #717, opened 2020-09-03
Disable gc.log logging by default

wixaw:
Hello

I propose this modification because these logs are useless and it's annoying to maintain a jvm.options on my side only for this parameter

thank you in advance


jmlrt:
Hi @wixaw, thanks for opening this PR.
The values in this files are matching default `jvm.options` file provided by Elasticsearch (see https://github.com/elastic/elasticsearch/blob/v7.9.1/distribution/src/config/jvm.options#L77).
We try to stay with default configuration as much as possible.

Issue #722, opened 2020-09-17
Collision of es_config/path.data/-log with es_data_dirs/-log_dir


vielfarbig:
**Elasticsearch version**
7.9.1
**Role version**:  (If using master please specify github sha)
commit a933f3c90ea18afe847648c58f1a48eec24d73ac
**JVM version** (`java -version`):
Default jdk ships with elastic:
```
/usr/share/elasticsearch/jdk/bin/java -version
openjdk version "14.0.1" 2020-04-14
OpenJDK Runtime Environment AdoptOpenJDK (build 14.0.1+7)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 14.0.1+7, mixed mode, sharing)
```
**OS version** (`uname -a` if on a Unix-like system):
Standard Ubuntu:
```
uname -a
Linux MACHINE_NAME 5.4.0-47-generic #51-Ubuntu SMP Fri Sep 4 19:50:52 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```
**Description of the problem including expected versus actual behaviour**:
If
```
...
  vars:
    es_config:
      path.data: "/opt/elasticsearch/data"
      path.logs: "/opt/elasticsearch/logs"
...
```
is set, instead of (if only this is set it works as expected)
```
...
  vars:
    es_data_dirs:
      - "/opt/elasticsearch/data"
    es_log_dir: "/opt/elasticsearch/logs"
    es_config:
...
```
`path.data` and `path.logs` are duplicated with the default path in the resulting `elasticsearch.yml`
```
...
path.data: /opt/elasticsearch/data
path.logs: /opt/elasticsearch/logs

#################################### Paths ####################################

# Path to directory containing configuration (this file and logging.yml):

path.data: /var/lib/elasticsearch

path.logs: /var/log/elasticsearch
...
``` 

Ansible fails with the error message below at the task "Check if bootstrap password is set", that seemingly has nothing to do with `path.*`-parsing, but I think, this case of `es_config/path.data/-log` should be handled explicitely while runnig the parsing task of the paths. 

**Playbook**:
```
- name: NAME
  hosts: master_node
  remote_user: USER
  become: yes
  roles:
    - role: elastic.elasticsearch
  vars:
    es_config:
      node.name: "NODE_NAME"
      path.data: "/opt/elasticsearch/data"
      path.logs: "/opt/elasticsearch/logs"
      cluster.name: "CLUSTER_NAME"
[...]

PR #725, opened 2020-10-02
Add note on only using es_data_dirs and es_log_dir for customizing

vielfarbig:
The es_data_dirs and es_log_dir leads to a duplication of the keys in `elasticsearch.yml` when using together with `es_config['path.data']` and `es_config['path.logs']`. For reference see https://github.com/elastic/ansible-elasticsearch/issues/722

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,10 +216,12 @@ A more complex example:
       - plugin: ingest-attachment
 ```
 
-#### Important Note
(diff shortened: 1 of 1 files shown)


vielfarbig:
I used wrong git name and email.
Sry this is my first PR on Github

PR #726, opened 2020-10-02
Add note to only using es_data_dirs and es_log_dir for customizing th…

vielfarbig:
…e data and log dirs (#722)

The es_data_dirs and es_log_dir leads to a duplication of the keys in `elasticsearch.yml` when using together with `es_config['path.data']` and `es_config['path.logs']`. 

Fix https://github.com/elastic/ansible-elasticsearch/issues/722

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,10 +216,12 @@ A more complex example:
       - plugin: ingest-attachment
 ```
 
-#### Important Note
+#### Important Notes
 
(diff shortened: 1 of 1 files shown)


jmlrt (approved):
LGTM⛴

jmlrt:
No need to run CI tests as there is only changes in the README.

jmlrt:
Thanks for this PR @vielfarbig 👍

PR #728, opened 2020-10-15
Improve the documentation for TLS

smutel:
Just add some part in the documentation to help user changing the password of default user (elastic).


jmlrt (approved):
LGTM⛴

PR #741, opened 2020-11-05
[meta] clean deprecated bumper script

jmlrt:
This commit remove the bumper script previously used to bump elasticsearch
version during releases. This scripts has been replaced by a fully
automated process available in Elastic private repos


nkammah (approved):
lgtm 🚢

PR #742, opened 2020-11-05
convert custom filter to python3

jmlrt:
This commit conver the custom filters to python 3 using 2to3

source: https://docs.python.org/2/library/2to3.html


Conky5 (approved):
LGTM

PR #685, opened 2020-03-30
Add es_keystore variable

Bernhard-Fluehmann:
Added variable to support adding and removing of elasticsearch.keystore entries

Fix #684

PR #757, opened 2021-01-21
Cherrypick es keystore entries

### LINKED ISSUE #684: Keystore Configuration
**Add suppot for elasticsearch.keystore entries**:
Some Elasticsearch features like watcher outputs or AD integration need entries stored in the elasticsearch keystore. As far as I can see this is currently not supported. Although the settings can be done outside of the playbook with a separate task it would still be easier to have this supported by the playbook.

**Elasticsearch version** 7.6.1

Bernhard-Fluehmann:
#Add support to add,update and delete  elasticsearch.keystore entries.
Obsoletes #685
Fix #684 

@bndabbs Please have a look


Bernhard-Fluehmann:
@jmlrt Could you add the feature label and assignement as from #685 to this PR? thanks

schallee:
Would it be better to use something like:

`- command:
    stdin: "{{passphrase}}"
    argv:
      - "{{keystore_cmd}}"
      - add
      - '-x'
      - '-f'
      - "xpack.security.{{ item }}.secure_password"
`

instead of echo pipe? I worry that certain values (eg: randomly generated password) could contain shell meta characters which would cause problems.

Bernhard-Fluehmann:
@schallee Thank yor for bringing this up. I have refactored it already and updated the pull request.

bndabbs:
Thanks for the contribution @BernardIsibor. I'm hoping to review this PR either today or tomorrow.

bndabbs (approved):
Looks good!

bndabbs:
jenkins test this please

tasks/xpack/security/elasticsearch-security.yml
```diff
@@ -28,12 +28,62 @@
 
   - name: Create Bootstrap password for elastic user
     become: yes
-    shell: echo {{ es_api_basic_auth_password | quote }} | {{ es_home }}/bin/elasticsearch-keystore add -x 'bootstrap.password'
+    command:
```
jmlrt, line 31: CI tests using this task are failing (example: https://devops-ci.elastic.co/job/elastic+ansible-elasticsearch+pull-request/237/OS=ubuntu-1404,TEST_TYPE=xpack-upgrade,label=linux/console).

I did a few local tests and see 2 issues:

- `argv` syntax for `command` module was released in Ansible 2.6 but our role is still supporting Ansible 2.5

  https://github.com/elastic/ansible-elasticsearch/blob/87946a8305147216cc682df35054f5e9b75ae953/meta/main.yml#L10

  I would prefer not remove 2.5 support if possible as it is the default version used in Ubuntu 18.04 packages and in our `ubuntu-1804` tests.

- The generated command don't seem valid. When using it with a more recent Ansible version and enabling log for this task, I get the following error:

  ```
  TASK [elasticsearch : Create Bootstrap password for elastic user] **************
  fatal: [localhost]: FAILED! => {"changed": false, "cmd": "changeme /usr/share/elasticsearch/bin/elasticsearch-keystore add -x bootstrap.password", "msg": "[Errno 2] No such file or directory", "rc": 2}
  ```

jmlrt (changes requested):

tasks/xpack/security/elasticsearch-security.yml
Bernhard-Fluehmann: Ok, I can have a look at this and change it back. How should we address the problem mentioned by @schalle about autogenerated passwords?

jmlrt: AFAIK the Ansible quote filter should handle the special characters correctly so that we don't have any issue.

I did a try with a few passwords like `elastic|$&01` and didn't have any issue with my small samples.

Bernhard-Fluehmann: Ok. Got it. I have fixed it and tried to push it but used a wrong git user. Can you delete this pull request please, I will do it again

Bernhard-Fluehmann: @jmlrt Or even better, just accept it.

jmlrt: Not sure that we can merge the PR with CLA failing on some commits.

Can you try changing the author name of these 2 commits with [this procedure](https://stackoverflow.com/a/3042512) for example?

Bernhard-Fluehmann: @jmlrt I have tried it. Can you please check? Thanks

jmlrt:
cla/check

tasks/xpack/security/elasticsearch-security.yml
jmlrt: This is still failing and I still see your other author name and emails in some commits. Then if you prefer creating a new branch/PR please do, you can then close this one or I'll do it when the new one will be created.

PR #760, opened 2021-01-26
[meta] add dedicated ci jobs for 6.x

jmlrt:
This commit create dedicated Jenkins jobs for 6.x for master branch and
PRs. This is required to allow having different test suites for 7.x and
6.x in a following PR.

PR #762, opened 2021-01-27
[meta] fix jobs template following #760

jmlrt:
This commit fix the test job templates following #760.
- add VERSION parameter for 7.x jobs
- move axis to jobs templates


mgreau (approved):
LGTM

PR #770, opened 2021-02-05
Remove OSS support for version >= 7.11.0

jmlrt:
This commit is removing the option to install Elasticsearch OSS
distribution for version >= 7.11.0.

This is due to Elasticsearch recent license change.
See https://www.elastic.co/blog/licensing-change for more details.

- Add notice to breaking changes and update documentation
- Fail the deployment if trying to install OSS >= 7.11.0
- Remove OSS tests for 7.x except the upgrade test from last OSS version
  to default distribution


mgreau (approved):
LGTM

PR #772, opened 2021-02-08
Added node.roles in readme and created deprecation warning

### LINKED ISSUE #731: Use node.roles instead of deprecated setting for Elasticsearch 7.9+
This role use `node.master` and `node.data` settings to define node roles.
However these settings have beed deprecated in favor of `node.roles` setting in 7.9 (https://github.com/elastic/elasticsearch/pull/54998).

Bernhard-Fluehmann:
Proposal to fix #731

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,11 @@ The following illustrates applying configuration parameters to an Elasticsearch 
       discovery.seed_hosts: "localhost:9301"
       http.port: 9201
       transport.port: 9301
-      node.data: false
-      node.master: true
+      node.roles:
+        - master
+        - ingest
+        - ml
+        - remote_cluster_client
       bootstrap.memory_lock: true
     es_heap_size: 1g
     es_api_port: 9201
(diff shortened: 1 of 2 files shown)


jmlrt:
Thanks for this PR.
This will need a bit more investigation in our side to ensure this is the way we want it to be implemented.

jmlrt (changes requested):
@Bernhard-Fluehmann, thanks for this PR and sorry for the delay.
I wanted to investigate if it would worth to do it another way to be consistent with how we'll handle this change in [Elasticsearch Helm chart](https://github.com/elastic/helm-charts/blob/master/elasticsearch/), however I think this PR is fine for Ansible Elasticsearch.

Can you also update: https://github.com/elastic/ansible-elasticsearch/blob/f0e4e360a2ceabc6e5bb0b9f73a808655bc875ea/docs/ssl-tls-setup.md#L65-L66

The rest LGTM:+1:

PR #777, opened 2021-02-19
Update ssl-tls-setup.md

DanRoscigno:
When generating the CA the filename is `my-ca.p12`, so I changed the name from `my-truststore.p12` to `my-ca.p12`


jmlrt (approved):
LGTM⛴

PR #779, opened 2021-02-22
Fix "list native roles" task

jmlrt:
This commit fix "List Native Roles" task with Elasticsearch 7.11.0+
by removing the unused [`body_format`](https://docs.ansible.com/ansible/2.9/modules/uri_module.html#parameter-body_format) parameter from `uri` module.

This parameter isn't used with a GET method but make request fail with
`request [GET /_security/role] does not support having a body` error
with Elasticsearch 7.11.0+.


Conky5 (approved):
LGTM

PR #787, opened 2021-03-16
Stop plugin install to fail in check mode

### LINKED ISSUE #786: Role execution fails if started in check mode
**Elasticsearch version**: `*`

**Role version**:  Tested with latest release `7.11.2` but also appears in previous versions

Role execution fails in Ansible check mode. It is expected that the role can also be executed fine when run with Ansible check mode activated.  
This happens in this task, especially in line `70/78`: https://github.com/elastic/ansible-elasticsearch/blob/4f01bc74a079c726a5045bed5595273fbe2d2bc0/tasks/elasticsearch-plugins.yml#L66-L80

Error message:

```
FAILED! => {"msg": "The conditional check 'plugin_installed.rc == 0' failed. The error was: error while evaluating conditional (plugin_installed.rc == 0): 'dict object' has no attribute 'rc'"}
```

This is due to the fact, that Ansible does not execute the command in check mode. This is why the registered variable does not have the `rc` parameter set and thus the role is currently failing when run in check mode.

tobiashuste:
In check mode Ansible is actually not going to execute the command from the `Install elasticsearch plugins` task.

https://github.com/elastic/ansible-elasticsearch/blob/4f01bc74a079c726a5045bed5595273fbe2d2bc0/tasks/elasticsearch-plugins.yml#L66-L80

Thus, the `rc` parameter is not set resulting in a failed execution. My suggestion would be to default the parameter to `0`, if it is not set. This allows the role to be run in check mode as well.

## Alternative solution

Alternatively, it would be possible to add

```yml
ignore_errors: "{{ ansible_check_mode }}"
```

to the task. But I thought the submitted solution might be cleaner.

Closes #786

### DIFF
--- a/tasks/elasticsearch-plugins.yml
+++ b/tasks/elasticsearch-plugins.yml
@@ -67,7 +67,7 @@
   become: yes
   command: "{{es_home}}/bin/elasticsearch-plugin install {{ item.url | default(item.plugin) }} --batch --silent"
   register: plugin_installed
-  changed_when: plugin_installed.rc == 0
+  changed_when: plugin_installed.rc|default(0) == 0
   with_items: "{{ es_plugins }}"
   when: item.plugin in plugins_to_install
   notify: restart elasticsearch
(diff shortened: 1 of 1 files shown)


jmlrt:
cla/check

jmlrt:
jenkins test this please

jmlrt:
Hi @tobiashuste,
Thanks for your PR.
Can you sign the CLA: https://www.elastic.co/contributor-agreement

tobiashuste:
@jmlrt Thanks, I just signed the CLA. I hope, it's all fine with it.

jmlrt (approved):
LGTM⛴

jmlrt:
jenkins test this please

PR #789, opened 2021-03-18
Use ES_JAVA_HOME instead of JAVA_HOME

jmlrt:
This commit set `ES_JAVA_HOME` environment variable instead of
`JAVA_HOME` when `es_java_home` environment variable is used.

See https://github.com/elastic/elasticsearch/issues/68848 for more
context.


Conky5 (approved):
LGTM

PR #796, opened 2021-05-20
Fails deployment when using tls without security

### LINKED ISSUE #756: Play does not fail when working with security and es_api_basic_auth_username and es_api_basic_auth_password are not provided
**Elasticsearch version** 7.10.1

**Role version**:  7.10.1

**JVM version** (`java -version`): `openjdk version "15.0.1" 2020-10-20`

**OS version** (`uname -a` if on a Unix-like system): CentOS 8 `Linux elk.example.com 4.18.0-240.1.1.el8_3.x86_64 #1 SMP Thu Nov 19 17:20:08 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

When specifying certificates for enabling HTTPS on Elasticsearch but not specifying any credentials the play should fail right away. It instead continues.
The expected behaviour is as exposed in [the comment of the check](https://github.com/elastic/ansible-elasticsearch/blob/ffeb286150bad5a6f6edbc3b805df24f2d159e93/tasks/elasticsearch-parameters.yml#L17) (`Check if working with security we have an es_api_basic_auth_username and es_api_basic_auth_username - otherwise any http calls wont work`) and should result in the following log:
```
TASK [elastic.elasticsearch : fail when api credentials are not declared when using security] ***
fatal: [192.168.128.87]: FAILED! => {"changed": false, "msg": "Enabling security requires an es_api_basic_auth_username and es_api_basic_auth_password to be provided to allow cluster operations"}
```

The certificates specified in the playbook exists on the ansible machine and are correctly uploaded on the system (because the play does not fail before).

**Playbook**:
Please specify the full playbook used to reproduce this issue.

```
- name: Configure Elasticsearch
  hosts: 192.168.128.87
  roles:
    - role: elastic.elasticsearch
  vars:
    es_config:
      node.name: "node1"
      cluster.name: "cluster"
      network.host: 192.168.128.87
      discovery.seed_hosts: [ 192.168.128.87 ]
      cluster.initial_master_nodes: [ 192.168.128.87 ]
    es_enable_http_ssl: true
    es_enable_transport_ssl: true
    es_ssl_certificate_authority: "files/ca/ca.crt"
    es_ssl_certificate: "files/elasticsearch/elasticsearch.crt"
    es_ssl_key: "files/elasticsearch/elasticsearch.key"
```

**Provide logs from Ansible**: extracted just the relevant part from the ansible play

```
TASK [elastic.elasticsearch : fail when api credentials are not declared when using security] ***
skipping: [192.168.128.87]
```

**ES Logs if relevant**: none

I experienced that changing the when clause of the check from
```
when:
    - not oss_version
    - es_api_basic_auth_username is defined
    - es_api_basic_auth_password is not defined
```
to
```
when:
    - not oss_version
    - es_api_basic_auth_username is not defined or es_api_basic_auth_password is not defined
```
would make the check as I intended it should work but maybe there is something more I'm missing.

jmlrt:
Fix #756

### DIFF
--- a/tasks/elasticsearch-parameters.yml
+++ b/tasks/elasticsearch-parameters.yml
@@ -36,13 +36,6 @@
     - (es_ssl_key == "" or es_ssl_certificate == "")
     - (es_ssl_keystore == "" or es_ssl_truststore == "")
 
-- name: fail when ssl enabled without defining a key and certificate
-  fail: msg="Enabling SSL/TLS (es_enable_http_ssl or es_enable_transport_ssl) requires es_ssl_keystore and es_ssl_truststore or es_ssl_key and es_ssl_certificate to be provided"
-  when:
-    - es_enable_http_ssl or es_enable_transport_ssl
-    - (es_ssl_key == "" or es_ssl_certificate == "")
-    - (es_ssl_keystore == "" or es_ssl_truststore == "")
-
 - name: set fact file_reserved_users
   set_fact: file_reserved_users={{ es_users.file.keys() | list | intersect (reserved_xpack_users) }}
   when: es_users is defined and es_users.file is defined and (es_users.file.keys() | list | length > 0) and (es_users.file.keys() | list | intersect (reserved_xpack_users) | length > 0)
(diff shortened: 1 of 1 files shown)


Conky5 (approved):
LGTM

PR #797, opened 2021-05-21
Fix versionlock centos 8

brad-x:
This resolves a corner case where a RHEL/CentOS 8 system does not have the `yum` compatibility package installed - everywhere else, the role is using the `yum` ansible module which detects and uses `dnf` in recent ansible versions, so the only concern is when:

- System is RHEL/CentOS 8
- System does not have `yum` installed and only `dnf` is available to the shell module
- Role variable `es_version_lock` is set to true, invoking the `elasticsearch-RedHat-version-lock.yml` playbook.

### DIFF
--- a/tasks/elasticsearch-RedHat-version-lock.yml
+++ b/tasks/elasticsearch-RedHat-version-lock.yml
@@ -2,13 +2,13 @@
 - name: RedHat - install yum-version-lock
   become: yes
   yum:
-    name: yum-plugin-versionlock
+    name: "{{ versionlock_plugin_package }}"
     state: present
     update_cache: yes
 
 - name: RedHat - check if requested elasticsearch version lock exists
   become: yes
-  shell: 'yum versionlock list | grep {{es_package_name}} | grep -c "{{es_version}}"'
+  shell: '{{ package_manager }} versionlock list | grep {{ es_package_name }} | grep -c "{{ es_version }}"'
   register: es_requested_version_locked
   args:
     warn: false
@@ -18,7 +18,7 @@
 
 - name: RedHat - lock elasticsearch version
   become: yes
-  shell: yum versionlock delete 0:elasticsearch* ; yum versionlock add {{ es_package_name }}-{{ es_version }}
+  command: '"{{ package_manager }}" versionlock delete 0:elasticsearch* ; "{{ package_manager }}" versionlock add {{ es_package_name }}-{{ es_version }}'
   args:
     warn: false
   when:
@@ -28,7 +28,7 @@
 
 - name: RedHat - check if any elasticsearch version lock exists
   become: yes
-  shell: yum versionlock list | grep -c elasticsearch
+  shell: '"{{ package_manager }}" versionlock list | grep -c elasticsearch'
   register: es_version_locked
   args:
     warn: false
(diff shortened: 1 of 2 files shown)


brad-x:
Apologies - I have now signed the contributor agreement. Does the PR need resubmission?

brad-x:
I see "CLA Expected - Waiting for status to be reported" - unsure if any further action is needed on my end after signing the CLA - please let me know if this is the case

tasks/elasticsearch-RedHat-version-lock.yml
```diff
@@ -2,13 +2,13 @@
 - name: RedHat - install yum-version-lock
```
darxriggs, line 2: `yum-version-lock` should also be replaced here with `{{ versionlock_plugin_package }}`.

tasks/elasticsearch-RedHat-version-lock.yml
```diff
@@ -2,13 +2,13 @@
...
   become: yes
   yum:
```
darxriggs, line 4: Should this then better be `package` to be more generic?

vars/RedHat.yml
```diff
@@ -2,3 +2,5 @@
 java: "{{ es_java | default('java-1.8.0-openjdk.x86_64') }}"
 default_file: "/etc/sysconfig/elasticsearch"
 es_home: "/usr/share/elasticsearch"
+package_manager: "{% if ansible_os_family == 'RedHat' and ansible_distribution_major_version >= '8' %}dnf{% else %}yum{% endif %}"
```
darxriggs, line 5: Better use `ansible_distribution_major_version | int >= 8` because for higher version numbers this comparison is not returning the expected result, e.g. `'10' >= '8'` evaluates to `false`.

darxriggs, line 5: `ansible_os_family` should not be required here because this file is only included in `tasks/main.yml` using `ansible_os_family` to derive the filename.

darxriggs:
Note: I am not a maintainer.

darxriggs:
Please keep it open.

vars/RedHat.yml
brad-x: Agreed. I primarily didn't want to diverge too far from existing statements in e.g. Debian.yml which has the same unnecessary check. But I'll remove it here as it's been discussed (I'm leaving it in Debian.yml for now though as I haven't tested there)

tasks/elasticsearch-RedHat-version-lock.yml
brad-x: Good call, fixed

tasks/elasticsearch-RedHat-version-lock.yml
brad-x: Also sensible

vars/RedHat.yml
brad-x: Done

brad-x: Actually I've just removed all that and gone with `ansible_facts.pkg_mgr` to simplify things.

brad-x:
Also I def signd the contributor agreement, about 3 times. Not sure why it thinks I haven't - possible issue with the commit workflow?

PR #823, opened 2021-10-02
Java 11 installation - Cover more OS versions

darxriggs:
- Debian based: Java 11 for all versions starting with Debian 10
- RedHat based: Java 11 for all versions starting with RHEL 7


darxriggs:
@jmlrt can you have a look at this?

PR #835, opened 2021-12-06
Remove Ubuntu 14.04 support

jmlrt:
This commit removes the support of Ubuntu 14.04. This distribution is
EOL since April 30th 2019 and CI tests are failing because apt
repositories aren't available anymore.


Conky5 (approved):
LGTM

